PR #10, opened 2022-07-12
Disable pm for join

peterharperuk:


dpgeorge:
Is this the right way to solve the problem?  Is there any precedence for this in the Broadcom/Cypress/Infineon driver?

I'm concerned that by disabling PM during join, it won't send the AP the correct DTIM values and the sleep mode won't work correctly when PM is enabled (after joining).  Did you test that power saving still works with this patch?

src/cyw43_ctrl.c
```diff
@@ -401,13 +405,22 @@ void cyw43_cb_process_async_event(void *cb_data, const cyw43_async_event_t *ev)
...
     if (self->wifi_join_state == WIFI_JOIN_STATE_ALL) {
+
         // STA connected
         self->wifi_join_state = WIFI_JOIN_STATE_ACTIVE;
         cyw43_cb_tcpip_set_link_up(self, CYW43_ITF_STA);
+        connect_done = true;
+    }
+
+    // Restore power management value
+    if (connect_done && self->saved_pm) {
```
dpgeorge, line 421: I think instead of `connect_done` you can check `(wifi_join_state & WIFI_JOIN_STATE_KIND_MASK) != 0`

src/cyw43_ctrl.c
```diff
@@ -401,13 +405,22 @@ void cyw43_cb_process_async_event(void *cb_data, const cyw43_async_event_t *ev)
...
+        cyw43_wifi_pm(self, self->saved_pm);
```
dpgeorge, line 422: I'm not sure it's safe to call this here.  Might need to do something like is done with `pend_rejoin` and set a flag to call this in `cyw43_poll_func()`.

src/cyw43_ctrl.c
peterharperuk: Won't that cause this line to get executed every poll - because WIFI_JOIN_STATE_ACTIVE is set?

peterharperuk:
> IIs there any precedence for this in the Broadcom/Cypress/Infineon driver?

I don't think there's a precedent, but I'm not sure the Infineon example will be very well tested.

> Did you test that power saving still works with this patch?

No - I just checked that the pm2_sleep_ret value was restored. I'll have to have a think about how to check if power management is affected.

src/cyw43_ctrl.c
peterharperuk: Actually - ignore that, saved_pm won't be set. I'll fix it as you suggest.

dpgeorge:
> No - I just checked that the pm2_sleep_ret value was restored. I'll have to have a think about how to check if power management is affected.

The way have done this in the past is to attach a current/power measuring device to the target board and set the DTIM high (eg 5, or even 9) and monitored the power usage over time when it's connected to the AP.  You need a relatively quiet WiFi network to see it go into low power mode.

src/cyw43_ctrl.c
peterharperuk: done

peterharperuk:
Apologies - I got confused a few times and pushed the wrong version. It should be correct now. I've done some testing. Disabling power management makes a big difference of around 8-10mA. I tried messing around with dtim and do see a small difference if I use a large value for dtim - maybe 0.5mA. As far as I can tell the power values are as expected after the pm is restored after a join so the change looks ok as far as I can tell.

kilograham:
I have a slight concern based on WiFi actually being down for a period, and that running batteries down.

I think maybe at least this should be optional (many don't have problems joining WiFi), and maybe should be periodically applied

peterharperuk:
The caller can set the pm value themselves to disable power management already. It would be useful if they had the function to query the value if we expected them to do that.

dpgeorge:
> I have a slight concern based on WiFi actually being down for a period, and that running batteries down.
> 
> I think maybe at least this should be optional (many don't have problems joining WiFi), and maybe should be periodically applied

This is a good point.

Maybe an alternative solution to disabling PM during join is to just increase the `pm2_sleep_ret` value?  Eg to it's maximum of 2000ms.  That would keep the 4343 awake for longer between periods of inactivity and might be all that is needed to successfully join the APs that are giving trouble.

Do we know which APs have trouble when PM is enabled?  If I can source one of these APs then I could test the fix.

peterharperuk:
Original report is here https://github.com/raspberrypi/pico-sdk/issues/912
So that's Zyxel VMG8825-T50

I believe I asked Floris (Maxnet) to test with 2000ms and it didn't seem to improve things.

dpgeorge:
> So that's Zyxel VMG8825-T50

It doesn't look like I can easily obtain one of those.

> I believe I asked Floris (Maxnet) to test with 2000ms and it didn't seem to improve things.

OK... Looking at some WICED examples, they do disable PM briefly during DHCP.  Also PM is not enabled until after connecting to WiFi, it seems.

PR #12, opened 2022-07-14
src: Event hooks during firmware load.

### LINKED ISSUE #11: Firmware loading seems to cause USB problems for macs
See https://github.com/micropython/micropython/issues/8904

We've tested a fix that adds strategic calls to MICROPY_EVENT_POLL_HOOK at appropriate places and it seems to fix the issue. We have to put some hooks in the driver.

peterharperuk:
Loading the wifi firmware can cause delays on startup which causes
problems for USB. Add some hooks so we can allow background processes to
run.

See https://github.com/micropython/micropython/issues/8904
and note that a change will be needed in micropython and pico-sdk to
define the hook.

Fixes #11

### DIFF
--- a/src/cyw43_ll.c
+++ b/src/cyw43_ll.c
@@ -466,6 +466,8 @@ static int cyw43_download_resource(cyw43_int_t *self, uint32_t addr, size_t raw_
     #endif
 
     for (size_t offset = 0; offset < len; offset += block_size) {
+        CYW43_EVENT_POLL_HOOK;
+
         size_t sz = block_size;
         if (offset + sz > len) {
             sz = len - offset;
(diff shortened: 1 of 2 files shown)


src/cyw43_ll.c
```diff
@@ -316,6 +316,11 @@ static const uint8_t wifi_nvram_4343[] __attribute__((aligned(4))) =
 static int cyw43_ll_sdpcm_poll_device(cyw43_int_t *self, size_t *len, uint8_t **buf);
 static void cyw43_write_iovar_n(cyw43_int_t *self, const char *var, size_t len, const void *buf, uint32_t iface);
 
+// This should be defined by the port, to let background processes like usb to run
+#ifndef CYW43_EVENT_POLL_HOOK
+#define CYW43_EVENT_POLL_HOOK
+#endif
+
```
dpgeorge, line 323: Please move this to `cyw43_config.h`.

src/cyw43_ll.c
```diff
@@ -466,6 +471,8 @@ static int cyw43_download_resource(cyw43_int_t *self, uint32_t addr, size_t raw_
     #endif
 
     for (size_t offset = 0; offset < len; offset += block_size) {
+        CYW43_EVENT_POLL_HOOK
```
dpgeorge, line 474: Please add a semicolon at the end of this line, to make it format like a statement.

peterharperuk: Done

src/cyw43_ll.c
peterharperuk: Done

peterharperuk:
@dpgeorge Is this ok now?

dpgeorge:
This is good now, thanks.  Merged in cd4075fc7d8c835431f5bed0e332f5a193990797

PR #13, opened 2022-07-15
rename FLASH_BLOCK_SIZE to CYW43_FLASH_BLOCK_SIZE 

kilograham:
This avoids a conflict with a Pico SDK header included by the rp2 port of MicroPython when using USB MSC. This is the easiest place to fix (SDK change is backwards incompatible; MP change is possible, but required duplicating a constant)


dpgeorge:
Thanks, that looks fine, it's indeed an invisible change (invisible to external user of this library).

Merged in 59f84b261a307161eaf3545037cf0005eb6e6f8e (with modification of commit message).

Issue #14, opened 2022-07-17
Allow user provided firmware and wifi_nvram_image


kilograham:
1. w.r.t. firmware, there is a `storage_read_blocks` function however it is not hooked up. if we could provide a callback (possibly enabled by `#define` to allow the driver to determine the size of the firmware and iterate over the blocks; this would help uses cases such as
    * https://github.com/raspberrypi/pico-sdk/issues/928 (sharing the firmware between multiple binaries)
    * https://github.com/raspberrypi/pico-sdk/issues/909 (compressing the firmware)

2. for completeness it seems like someone might want to tweak the wifi_nvram_image, which could be done by just allowing them to inject a #define.

Note right now both of these pieces of code are somewhat peculiar to the bus transport; idk if this should be done for both SPI and SDIO (I don't see why not though)

dpgeorge:
See #15 for the part (2), to make wifi_nvram configurable.

peterharperuk:
Has this been fixed?

dpgeorge:
In 82d2037f6af74d585ed1f9cec9f4e8166c4e84b0 the firmware blob was changed to a header file.  So now the config value `CYW43_CHIPSET_FIRMWARE_INCLUDE_FILE` can be used to set which firmware to include.

PR #15, opened 2022-07-19
firmware: Make wifi_nvram settings configurable.

dpgeorge:
A port can define CYW43_WIFI_NVRAM_INCLUDE_FILE to select this config file.

Addresses part of issue #14.


dpgeorge:
@kilograham does this look reasonable to you?  It allows the user to define the include file `CYW43_WIFI_NVRAM_INCLUDE_FILE` if needed.

The alternative would be to put the nvram settings in a binary blob like the firmware and CLM.  Then it could be compressed as well.

kilograham (approved):
lgtm

PR #22, opened 2022-08-20
Fix LWIP mDNS TTL bug

siggimoo:
LWIP [fixed an mDNS bug](https://github.com/lwip-tcpip/lwip/commit/62fb2fd749ba0bf5fed8c494e840075663484347) that resulted in the removal of RR TTL parameters. This change brings the repo into sync with that.


dpgeorge:
Thanks for the contribution.

mDNS support was removed from this driver in df2c2e2af32bb01f8441fe58f735230ab14af83d, so this PR is no longer relevant.

PR #24, opened 2022-08-25
Implement a multicast filtering api and enable v6 for lwip

felixdoerre:
This is related to #19 should resolve #18.
Thanks to the good hint in from @peterharperuk  https://github.com/georgerobotics/cyw43-driver/issues/18#issuecomment-1225913653 I was able to use the mac address filtering. With these changes applied, ipv6 works perfectly for me.
And hints? Should I split up the 2 commits into to separate PRs?

PR #25, opened 2022-08-26
src: Implement LWIP's IGMP MAC filter callback.

jimmo:
This is used to register the multicast group address on the wifi
interface's filter, allowing frames to that group to be received.

This only applies to v4. See #24 for the equivalent v6 change.

Fixes https://github.com/micropython/micropython/issues/9105

### DIFF
--- a/src/cyw43_lwip.c
+++ b/src/cyw43_lwip.c
@@ -39,6 +39,7 @@
 #if CYW43_LWIP
 #include "lwip/etharp.h"
 #include "lwip/dns.h"
+#include "lwip/igmp.h"
 #include "lwip/apps/mdns.h"
 #include "lwip/tcpip.h"
 #include "netif/ethernet.h"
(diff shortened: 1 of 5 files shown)


jimmo:
Note:
- Unsure if the `cyw43_delay_ms(50);` is needed at the end of `cyw43_ll_wifi_update_multicast_filter`.
- LWIP manages the group memberships, so in theory `cyw43_ll_wifi_update_multicast_filter` doesn't need to validate if a group is already registered. It's not much code though to ensure that they don't get out of sync.

felixdoerre:
this change and #24 have quite a bit in common. I've split my changes into two commits, and I think [2baa5c5](https://github.com/georgerobotics/cyw43-driver/pull/24/commits/2baa5c589797f68954d1248b1179cb7d5f771c46) can be used directly here, I think we should agree on a common interface to set the mac addresses, so mcast registered for v4 does not conflict with addresses registered for v6. I left out the 50ms delay and it worked fine on my board, but that might be something that someone more familiar with the chip can decide. You also removed the hardcoded mcast-address for mdns. I am not sure if that is needed, but I imagine it was in there for a reason.

jimmo:
> You also removed the hardcoded mcast-address for mdns. I am not sure if that is needed, but I imagine it was in there for a reason.

LWIP invokes the callback with the mdns group address when the mdns hostname is configured. The existing hard coded address was a workaround for not implementing this callback.

> I think [2baa5c5](https://github.com/georgerobotics/cyw43-driver/pull/24/commits/2baa5c589797f68954d1248b1179cb7d5f771c46) can be used directly here, I think we should agree on a common interface to set the mac addresses, so mcast registered for v4 does not conflict with addresses registered for v6.

This PR should replace that commit. It provides exactly the same functionality just without needing the extra 60 bytes of RAM to manage the list.

And to my first note above, using the same method for v6 is a good reason to keep the code to track add duplicate detection in cyw43_ll, as I think LWIP can't do this across v4&v6.

felixdoerre:
sounds good, I'll rebase the v6 PR on this one.

peterharperuk:
This change is now in https://github.com/georgerobotics/cyw43-driver/pull/24

dpgeorge:
Thanks, this is a good improvement, definitely needed!

src/cyw43_lwip.c
```diff
@@ -100,13 +101,29 @@ STATIC err_t cyw43_netif_output(struct netif *netif, struct pbuf *p) {
     return ERR_OK;
 }
 
+STATIC err_t cyw43_netif_update_igmp_mac_filter(struct netif *netif, const ip4_addr_t *group, enum netif_mac_filter_action action) {
```
dpgeorge, line 104: this was originally cyw32; fixed during merged

Issue #17, opened 2022-07-23
Uses outdated lwIP API?


lurch:
Related to https://github.com/raspberrypi/pico-examples/issues/247

Whilst I wasn't able to reproduce the originally-reported problem in that issue, curiosity prompted me to do a bit of digging anyway... :wink: 

https://github.com/georgerobotics/cyw43-driver/blob/main/src/cyw43_lwip.c#L182 does
```
mdns_resp_add_netif(n, mdns_hostname, 60);
```
however the third parameter to `mdns_resp_add_netif` was removed from lwIP on [13th November 2018](https://github.com/lwip-tcpip/lwip/commit/62fb2fd749ba0bf5fed8c494e840075663484347). Looking at the dates of the [lwIP release-tags](https://github.com/lwip-tcpip/lwip/tags), I guess this means that there was a backwards-incompatible API change between the 2.1.1 and 2.1.2 versions of lwIP :confused:  And annoyingly, even though https://savannah.nongnu.org/projects/lwip mentions the 2.1.3 release of lwIP (which agrees with GitHub), http://www.nongnu.org/lwip/ shows the Doxygen for the 2.1.0 release, which uses the [old API](https://www.nongnu.org/lwip/2_1_x/group__mdns.html#gaa619ac8f46a4b4021195720f0355cbeb) :man_facepalming:

dpgeorge:
In MicroPython we use the `STABLE-2_1_2_RELEASE` of lwIP.  The signature of the MDNS function in that release is:
```c
err_t mdns_resp_add_netif(struct netif *netif, const char *hostname, u32_t dns_ttl);
```
So it looks correct for that version.

I guess the point is, we should update to `STABLE-2_1_3_RELEASE` :smile:

lurch:
> In MicroPython we use the STABLE-2_1_2_RELEASE of lwIP.

Well that's kinda awkward, as [pico-sdk](https://github.com/raspberrypi/pico-sdk/tree/master/lib) uses `lwip @ 239918c` which appears to be the [current HEAD](https://github.com/lwip-tcpip/lwip/tree/master) of lwIP `master`.
ping @kilograham and @peterharperuk 

Does the code in `cyw43-driver` need to be conditional on what version of lwIP it's being compiled against?

[...]

PR #30, opened 2022-09-08
src: mDns fixes.

### LINKED ISSUE #16: Allow user to specify hostname
Currently `CYW43_HOST_NAME` can be specified at compile time, and is used here https://github.com/georgerobotics/cyw43-driver/blob/cd4075fc7d8c835431f5bed0e332f5a193990797/src/cyw43_lwip.c#L158

I'm not sure if you can call `netif_set_hostname` after this method returns and have it take effect (i'm guessing not?)

If not, it would probably nice to put an un-owned `const char *` in driver state, which if NULL uses `CYW43_HOST_NAME` (alternatively it can just default to that).

The main use case for this is actually more from MicroPython than C where I've seen people asking for it (in C it is easy enough to just override the #define).

In either case, there is also this line, the purpose of which I don't fully understand

https://github.com/georgerobotics/cyw43-driver/blob/195dfcc10bb6f379e3dea45147590db2203d3c7b/src/cyw43_lwip.c#L179

peterharperuk:
The host name for mDns wasn't using CYW43_HOST_NAME. The mdns_resp_add_netif changed after lwip version 2.1.2. Avoid calling mdns_resp_remove_netif if it's not been added. Call mdns_resp_announce when STA connected.

Fixes #16 and #17

### DIFF
--- a/src/cyw43.h
+++ b/src/cyw43.h
@@ -116,6 +116,7 @@ typedef struct _cyw43_t {
     // lwIP data
     struct netif netif[2];
     struct dhcp dhcp_client;
+    bool mdns_init;
     #endif
 
     #if CYW43_NETUTILS
--- a/src/cyw43_ctrl.c
+++ b/src/cyw43_ctrl.c
@@ -47,6 +47,10 @@
 #define USE_SDIOIT (1)
 #endif
 
+#if CYW43_LWIP
+#include "lwip/apps/mdns.h"
+#endif
+
 // Bits 0-3 are an enumeration, while bits 8-11 are flags.
 #define WIFI_JOIN_STATE_KIND_MASK (0x000f)
 #define WIFI_JOIN_STATE_ACTIVE  (0x0001)
@@ -408,6 +412,10 @@ void cyw43_cb_process_async_event(void *cb_data, const cyw43_async_event_t *ev) 
         // STA connected
         self->wifi_join_state = WIFI_JOIN_STATE_ACTIVE;
         cyw43_cb_tcpip_set_link_up(self, CYW43_ITF_STA);
+
+        #if LWIP_MDNS_RESPONDER
+        mdns_resp_announce(&self->netif[CYW43_ITF_STA]);
+        #endif
     }
 }
 
--- a/src/cyw43_lwip.c
+++ b/src/cyw43_lwip.c
@@ -42,6 +42,7 @@
 #include "lwip/apps/mdns.h"
 #include "lwip/tcpip.h"
 #include "netif/ethernet.h"
+#include "lwip/init.h"
 #endif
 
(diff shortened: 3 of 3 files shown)


lurch:
> The mdns_resp_add_netif changed after lwip version 2.1.2

I don't think this is right?
https://github.com/lwip-tcpip/lwip/blob/STABLE-2_1_2_RELEASE/src/include/lwip/apps/mdns.h#L76 says
```c
err_t mdns_resp_add_netif(struct netif *netif, const char *hostname, u32_t dns_ttl);
```
and https://github.com/lwip-tcpip/lwip/blob/STABLE-2_1_3_RELEASE/src/include/lwip/apps/mdns.h#L76 also says
```c
err_t mdns_resp_add_netif(struct netif *netif, const char *hostname, u32_t dns_ttl);
```

It's only in https://github.com/lwip-tcpip/lwip/blob/master/src/include/lwip/apps/mdns.h#L110 where it changes to
```c
err_t mdns_resp_add_netif(struct netif *netif, const char *hostname);
```

Looking at the [branch names](https://github.com/lwip-tcpip/lwip/branches) they have `master` and `STABLE-2_1_x` so perhaps the appropriate fix here is to check against `LWIP_VERSION_MAJOR` and `LWIP_VERSION_MINOR` from https://github.com/lwip-tcpip/lwip/blob/STABLE-2_1_x/src/include/lwip/init.h ? :man_shrugging: 

https://github.com/lwip-tcpip/lwip/blob/master/src/include/lwip/init.h seems to suggest that the current `master` will eventually become 2.2.x ?

dpgeorge:
Thanks for the patch.

Honestly, I think it might be better to just remove all mDNS code from this driver and push that responsibility to the user of this driver.  There are many ways that mDNS could be used/configured and it doesn't really make sense to add lots of configurability to this driver to support setting the mDNS in the way the application needs/wants.

We can instead have hooks/callbacks in this driver at various points (eg where the existing call to `mdns_resp_add_netif` is) to facilitate starting up lwIP (and other) components.

Would that work for pico-sdk?  Pico-sdk could provide some convenience functions to enable mDNS, or the user could just do it all themselves.

Note also that there are subtly different functions to use to announce/restart mDNS (from the lwIP docs):
- Call mdns_resp_announce() every time the IP address on the netif has changed.
- Call mdns_resp_restart() every time the network interface comes up after being
down, for example cable connected after being disconnected, administrative
interface comes up after being down, or the device wakes up from sleep.

There's also the config option `MDNS_RESP_USENETIF_EXTCALLBACK` which seems very useful, it calls these announce/restart functions automatically.  That's the kind of option the user could enable.

peterharperuk:
Ok. I imagine it should be possible to do this in the sdk. I'll have to study it. Let me close this for now.

PR #42, opened 2022-10-18
Support user-supplied CYW43_FW_DATA_ADDR

esky-software:
Allows the firmware blob to be provided by external means (e.g. loaded to a fixed location in flash).

Also add compile guards to cyw43_sdio.c, so that the file doesn't have to be excluded from the build if CYW43_USE_SPI is defined.


peterharperuk:
Looks ok to me. I'm working on a bigger change that should in theory allow you to load the firmware from flash and uncompress it.

dpgeorge:
You can now override the firmware blob by defining `CYW43_CHIPSET_FIRMWARE_INCLUDE_FILE`.

PR #44, opened 2022-11-22
Change firmware

peterharperuk:
Refactor the driver to allow firmware compression


peterharperuk:
Don't know what happened there! Re-opened pull request.

dpgeorge:
It should be relatively straightforward to take the compressed binary blob and run it through `xxd -i`?

Issue #27, opened 2022-08-29
Open (unsecured) APs are not supported


adamkaplan:
Referencing this MicroPython issue https://github.com/micropython/micropython/issues/9016#issuecomment-1219896195

Password-less AP support is disabled, [ostensibly by mistake](https://github.com/georgerobotics/cyw43-driver/blob/2ab6ca93f9cd044bc6f35c1403b1284e4161294a/src/cyw43_ll.c#L2018). However, even if that is fixed, the 802.11 security event processing doesn't appear to support APs with security disabled.

Clarke-Jools-36:
I have replicated this issue on MicroPython v1.19.1-88-g74e33e714, any updates on a fix?

gwo0d:
Likewise, I have replicated this.

gwo0d:
Submitted a patch for this here: #45

dpgeorge:
Fixed in commits 61432f360748949ad06564193a93d9fdb5afe9dd and 2cf328d9e41603405a037a29e081a7d30dd519e6 (it can now connect with CYW43_AUTH_OPEN).

PR #45, opened 2022-11-23
Add handling for open networks to cyw43_ll_wifi_join.

gwo0d:
Fix for #27 
- Add handling for auth_type == 0 to IF.
- The IF on line 2038 provided handling for networks without security (open networks). However, this handling is never used due to it being ignored by the IF on line 2013.
- Reproducibly fixes open network connection issues observed by Raspberry Pi Pico W MicroPython users (discussed [here](https://github.com/micropython/micropython/issues/9016#issuecomment-1219896195)) using `security=0` parameter.

### DIFF
--- a/src/cyw43_ll.c
+++ b/src/cyw43_ll.c
@@ -2055,7 +2055,9 @@ int cyw43_ll_wifi_join(cyw43_ll_t *self_in, size_t ssid_len, const uint8_t *ssid
     }
 
     uint32_t wpa_auth = 0;
-    if (auth_type == CYW43_AUTH_WPA2_AES_PSK || auth_type == CYW43_AUTH_WPA2_MIXED_PSK) {
+    if (auth_type == CYW43_AUTH_OPEN) {
+        wpa_auth = 0;
+    } else if (auth_type == CYW43_AUTH_WPA2_AES_PSK || auth_type == CYW43_AUTH_WPA2_MIXED_PSK) {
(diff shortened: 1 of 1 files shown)


peterharperuk (commented):
Looks good

src/cyw43_ll.c
```diff
@@ -2014,6 +2014,8 @@ int cyw43_ll_wifi_join(cyw43_ll_t *self_in, size_t ssid_len, const uint8_t *ssid
         wpa_auth = CYW43_WPA2_AUTH_PSK;
     } else if (auth_type == CYW43_AUTH_WPA_TKIP_PSK) {
         wpa_auth = CYW43_WPA_AUTH_PSK;
+    } else if (auth_type == 0) {
+        wpa_auth = CYW43_AUTH_OPEN;
```
dpgeorge, line 2018: I think the constants should be the other way around, ie:
```c
} else if (auth_type == CYW43_AUTH_OPEN) {
    wpa_auth = 0;
```
That will be consistent with the other if conditions here, checking against `CYW43_AUTH_xxx` constants.

gwo0d: This seems good to me.

dpgeorge:
Thanks for updating.  I also moved the new "if" up to the start of this set of if statements, because, logically, handling the case of open network should come before handling a network with security.

PR #46, opened 2022-11-23
Added country codes for all (most) European countries

Ioan-Cristian:
Added country codes for all (most) European countries.
@kilograham @peterharperuk


peterharperuk:
Do they work? The current list was compiled from the countries supported by the wifi firmware

PR #49, opened 2022-12-05
Fix ip address config

peterharperuk:
A couple of fixes for issues related to IP addresses and disabling DHCP


src/cyw43_lwip.c
```diff
@@ -124,6 +124,31 @@ STATIC err_t cyw43_netif_init(struct netif *netif) {
...
+#define CYW43_DEFAULT_IP_GATEWAY LWIP_MAKEU32(192, 168, 0, 1)
+#endif
+
+#ifndef CYW43_DEFAULT_IP_DNS
+#define CYW43_DEFAULT_IP_DNS LWIP_MAKEU32(8, 8, 8, 8)
+#endif
+
+#ifndef CYW43_DEFAULT_AP_IP_ADDRESS
+#define CYW43_DEFAULT_AP_IP_ADDRESS LWIP_MAKEU32(192, 168, 4, 1)
+#endif
```
dpgeorge, line 150: Can you please move these to `cyw43_config.h`, to keep all config in one place?

peterharperuk: Done

peterharperuk:
Rebase done. The only slight issue is you have to define the static IP address as a number, e.g. CYW43_DEFAULT_IP_ADDRESS=0x0a030fe8. With a bit more work I could make it more friendly, or use some cmake magic?

src/cyw43_config.h
```diff
@@ -115,3 +115,31 @@
 #ifndef CYW43_EVENT_POLL_HOOK
 #define CYW43_EVENT_POLL_HOOK
 #endif
+
+#if LWIP_DHCP
+// If LWIP_DHCP is true, the default IP address has to be zero for isconnected()
+// to know when DHCP has assigned an IP address
+// If LWIP_DHCP is false, we expect CYW43_DEFAULT_IP_ADDRESS to be the set to
```
lurch, line 122: "be the set to" -> "be set to"

src/cyw43_config.h
```diff
@@ -115,3 +115,31 @@
...
+#ifndef CYW43_DEFAULT_IP_GATEWAY
+#define CYW43_DEFAULT_IP_GATEWAY LWIP_MAKEU32(192, 168, 0, 1)
+#endif
+
+#ifndef CYW43_DEFAULT_IP_DNS
+#define CYW43_DEFAULT_IP_DNS LWIP_MAKEU32(8, 8, 8, 8)
+#endif
+
+#ifndef CYW43_DEFAULT_AP_IP_ADDRESS
+#define CYW43_DEFAULT_AP_IP_ADDRESS LWIP_MAKEU32(192, 168, 4, 1)
```
lurch, line 144: Is it "correct" that `CYW43_DEFAULT_IP_GATEWAY` is in the `192.168.0.*` network, but `CYW43_DEFAULT_AP_IP_ADDRESS` is in the `192.168.4.*` network?

peterharperuk: I just copied the values that were in the code already. I'll have to check. It doesn't look entirely correct when the mask is 255.255.255.0

src/cyw43_config.h
```diff
@@ -115,3 +115,31 @@
 #ifndef CYW43_EVENT_POLL_HOOK
 #define CYW43_EVENT_POLL_HOOK
 #endif
+
+#if LWIP_DHCP
```
dpgeorge, line 119: I don't think this will work because lwip headers are not included until after this header files, so this define will not be set.

Probably best to just define `CYW43_DEFAULT_IP_ADDRESS` here as all other IP settings are defined, then in `cyw43_lwip.c` have something like:
```c
        #if LWIP_DHCP
        IP4_ADDR(&IP(ipconfig[0]), 0, 0, 0, 0);
        #else
        IP(ipconfig[0]).addr = PP_HTONL(CYW43_DEFAULT_IP_ADDRESS);
        #endif
```

src/cyw43_config.h
dpgeorge: It's correct as-is: there's a difference between STA and AP addresses.

In fact I think it would be best to rename the constants to the following, to make it clearer:
```c
CYW43_DEFAULT_IP_STA_ADDRESS
CYW43_DEFAULT_IP_STA_GATEWAY
CYW43_DEFAULT_IP_AP_ADDRESS
CYW43_DEFAULT_IP_AP_GATEWAY
```
(the last two will have the same value of 192.168.4.1 by default)

dpgeorge:
> The only slight issue is you have to define the static IP address as a number

I think that's fine, one can just use `LWIP_MAKEU32()` like you have done in the default config.

src/cyw43_config.h
peterharperuk: Hopefully fixed as suggested. I left mask and dns alone as they're common between sta and ap.

peterharperuk:
> one can just use LWIP_MAKEU32() like you have done in the default config

I would set the static ip address in the cmake file not in any source code. For instance, I put this in some test code to test the change...
CYW43_DEFAULT_IP_STA_ADDRESS=0x0a030f19
Not terrible but not that obvious?

lurch:
I wonder if https://cmake.org/cmake/help/latest/command/math.html is sophisticated enough to allow you to write an "IP address to hexadecimal" function/macro directly in CMake? :shrug:

peterharperuk:
It might be easier to just make the code accept a value in the form a.b.c.d. It would have to convert it to a string and then convert the string to an IP address. It's just added complication that could go wrong if there's a formatting issue.

lurch:
I've never written any CMake myself, but I was imagining something like `IP_OCTETS_TO_HEX(192, 168, 0, 1)` and ignoring the string-parsing part entirely :wink:

peterharperuk:
@dpgeorge I've just rebased this. Are you happy with it?

peterharperuk:
I could add a function like IP_OCTETS_TO_HEX to the sdk and update one of the examples to show how it's used. But I think this driver change is enough.

peterharperuk:
https://github.com/peterharperuk/pico-sdk/commit/5d4ee66a3d80dd11283375b41de669b009dd43a8

src/cyw43_config.h
dpgeorge: @peterharperuk I think this comment is still relevant, and my suggestion is still the way I think it should be done.

src/cyw43_config.h
dpgeorge: @peterharperuk there should be separate `AP_ADDRESS` and `AP_GATEWAY` (although default to the same value).

Also, did you want to name the STA specific ones `CYW43_DEFAULT_IP_STA_ADDRESS` and `CYW43_DEFAULT_IP_STA_GATEWAY`?

dpgeorge:
@peterharperuk let me know what you think about the above comments of mine.  I'm happy to make the changes, just let me know either way.

peterharperuk:
Sorry was doing something else, the comments seem sensible as usual :) I'll push the changes suggested

dpgeorge:
Great!  There's no real hurry here.

src/cyw43_config.h
peterharperuk: Apologies, you're quite right. Should be fixed as suggested now

src/cyw43_config.h
peterharperuk: added AP_ADDRESS and AP_GATEWAY and put STA in the other defines.

dpgeorge:
Thanks for updating.  Now merged.

PR #50, opened 2022-12-05
src: Add a function to get RSSI.

### LINKED ISSUE #35: RFE: an API call to get the RSSI of just the currently connected AP
Related to #34: I suggest a function in the API that returns the RSSI of the currently connected AP.

In #34 I described running `cyw43_wifi_scan()` (now working), which I was doing because it is apparently the only way to get any RSSI value. In fact I just want to find it out for the current AP, so I'm running the scan and getting the RSSI in the callback when the SSID matches, and ignoring every other result of the scan.

In [this forum thread](https://forums.raspberrypi.com/viewtopic.php?t=340680), a Raspberry Pi engineer stated that obtaining the RSSI for the current AP is supported, but not exposed by the driver. So I think it would be a sensible addition (and would avoid a lot of superfluous scanning).

peterharperuk:
Fixes #35

### DIFF
--- a/src/cyw43.h
+++ b/src/cyw43.h
@@ -337,6 +337,18 @@ int cyw43_wifi_join(cyw43_t *self, size_t ssid_len, const uint8_t *ssid, size_t 
 int cyw43_wifi_leave(cyw43_t *self, int itf);
 
 /*!
+ * \brief Get the signal strength (RSSI) of the wifi network
+ *
+ * For STA (client) mode, returns the signal strength or RSSI of the wifi network.
+ * An RSSI value of zero is returned if you call this function before a network is connected.
+ *
+ * \param self the driver state object. This should always be \c &cyw43_state
+ * \param rssi a pointer to which the returned RSSI value is stored.
+ * \return 0 on success
+ */
+int cyw43_wifi_get_rssi(cyw43_t *self, int32_t *rssi);
+
+/*!
  * \brief Get the ssid for the access point
  *
  * For access point (AP) mode, this method can be used to get the SSID name of the wifi access point.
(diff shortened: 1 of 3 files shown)


src/cyw43.h
```diff
@@ -321,6 +321,17 @@ int cyw43_wifi_join(cyw43_t *self, size_t ssid_len, const uint8_t *ssid, size_t
...
+/*!
+ * \brief Get the signal strength (RSSI) of the wifi network
+ *
+ * For STA (client) mode, returns the signal strength or RSSI of the wifi network.
+ * An RSSI value of zero is returned if you call this function before a network is connected.
+ *
+ * \param self the driver state object. This should always be \c &cyw43_state
+ * \return 0 on success
+ */
+int cyw43_get_rssi(cyw43_t *self, int *rssi);
```
dpgeorge, line 333: I think this should be called `cyw43_wifi_get_rssi()`.  And the `rssi` argument should be `int32_t` to be clear that it's 32-bit.

peterharperuk: Rebased and fixed.

src/cyw43.h
```diff
@@ -336,6 +336,17 @@ int cyw43_wifi_join(cyw43_t *self, size_t ssid_len, const uint8_t *ssid, size_t
...
 int cyw43_wifi_leave(cyw43_t *self, int itf);
 
+/*!
+ * \brief Get the signal strength (RSSI) of the wifi network
+ *
+ * For STA (client) mode, returns the signal strength or RSSI of the wifi network.
+ * An RSSI value of zero is returned if you call this function before a network is connected.
+ *
+ * \param self the driver state object. This should always be \c &cyw43_state
+ * \return 0 on success
```
lurch, line 346: Should there be a `\param rssi` here too?

dpgeorge: I added this during the merge.

PR #52, opened 2022-12-18
Make cyw43_async_event_name_table const so it goes to flash

bruelltuete:
Checking the map file, `cyw43_async_event_name_table` was in RAM instead of flash, despite being "read only".
Needs an additional const.

### DIFF
--- a/src/cyw43_ctrl.c
+++ b/src/cyw43_ctrl.c
@@ -272,7 +272,7 @@ void cyw43_cb_ensure_awake(void *cb_data) {
     #endif
 }
 
-STATIC const char *cyw43_async_event_name_table[89] = {
+STATIC const char *const cyw43_async_event_name_table[89] = {
     //[0 ... 88] = NULL,
     [CYW43_EV_SET_SSID] = "SET_SSID",
     [CYW43_EV_JOIN] = "JOIN",


dpgeorge:
I checked a few builds and this table does actually go in ROM (text/rodata sections) because the compiler sees it's static and sees that it's not written to, so optimises it into ROM.

But you're right that it should have a const, to make the intention clear and make sure it does always go into the text or rodata section.

dpgeorge:
Thanks for the contribution!

bruelltuete:
> I checked a few builds and this table does actually go in ROM (text/rodata sections) because the compiler sees it's static and sees that it's not written to, so optimises it into ROM.

For debug builds it doesn't automatically go to read-only. The linker map says:
```
 .data.cyw43_async_event_name_table
                0x0000000020001628      0x164 CMakeFiles/farty.dir/pico-sdk/lib/cyw43-driver/src/cyw43_ctrl.c.obj
```

And from the release build:
```
 .rodata.cyw43_async_event_name_table
                0x000000001001f374      0x164 CMakeFiles/farty.dir/pico-sdk/lib/cyw43-driver/src/cyw43_ctrl.c.obj
```

With the extra const it goes to .rodata in both cases.
this is for a raspi pico project... i don't know if that's different for other targets...

dpgeorge:
Ah, OK, that makes sense that it wouldn't optimise it into ROM for debug builds.

(Note that this fix was merged.)

PR #56, opened 2023-01-24
src: Channel fix in cyw43_ll_wifi_join.

### LINKED ISSUE #54: cyw43_ll_wifi_join doesn't work with a BSSID
It's setting the channel count to 1 even when no channel is supplied. This seems to stop BSSID selection working. See https://github.com/raspberrypi/pico-sdk/pull/1091

peterharperuk:
Joining to wifi via BSSID doesn't appear to work.
Don't set the channel count if the channel is not set.

Fixes #54

### DIFF
--- a/src/cyw43_ll.c
+++ b/src/cyw43_ll.c
@@ -2129,9 +2129,11 @@ int cyw43_ll_wifi_join(cyw43_ll_t *self_in, size_t ssid_len, const uint8_t *ssid
         #define WL_CHANSPEC_CTL_SB_NONE     WL_CHANSPEC_CTL_SB_LLL
         #define WL_CHANSPEC_BAND_2G        0x0000
         memcpy(buf + 4 + 32 + 20, bssid, 6);
-        cyw43_put_le32(buf + 4 + 32 + 20 + 8, 1); // chanspec_num
-        uint16_t chspec = channel | WL_CHANSPEC_BW_20 | WL_CHANSPEC_CTL_SB_NONE | WL_CHANSPEC_BAND_2G;
-        cyw43_put_le16(buf + 4 + 32 + 20 + 12, chspec); // chanspec_list
+        if (channel != CYW43_CHANNEL_NONE) {
+            cyw43_put_le32(buf + 4 + 32 + 20 + 8, 1); // chanspec_num
+            uint16_t chspec = channel | WL_CHANSPEC_BW_20 | WL_CHANSPEC_CTL_SB_NONE | WL_CHANSPEC_BAND_2G;
+            cyw43_put_le16(buf + 4 + 32 + 20 + 12, chspec); // chanspec_list
+        }
 
         // join the AP
         CYW43_VDEBUG("Join AP\n");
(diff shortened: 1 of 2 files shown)


dpgeorge:
Thanks, this looks OK, but I'll have to do some testing.

I have used this feature in the past (join to a specific BSSID), but I see now that in MicroPython it passes through `channel=-1` as the default, and maybe that was just working by accident.

peterharperuk:
I couldn't seem to get it to work in Micropython where -1 is used. I needed to apply the fix and pass zero for the channel. If I also check for -1 it makes Micropython work when no channel is specified. I'll push that.

src/cyw43_ll.c
```diff
@@ -2127,9 +2127,11 @@ int cyw43_ll_wifi_join(cyw43_ll_t *self_in, size_t ssid_len, const uint8_t *ssid
         #define WL_CHANSPEC_CTL_SB_NONE     WL_CHANSPEC_CTL_SB_LLL
         #define WL_CHANSPEC_BAND_2G        0x0000
         memcpy(buf + 4 + 32 + 20, bssid, 6);
-        cyw43_put_le32(buf + 4 + 32 + 20 + 8, 1); // chanspec_num
-        uint16_t chspec = channel | WL_CHANSPEC_BW_20 | WL_CHANSPEC_CTL_SB_NONE | WL_CHANSPEC_BAND_2G;
-        cyw43_put_le16(buf + 4 + 32 + 20 + 12, chspec); // chanspec_list
+        if (channel != 0 && channel != -1) {
```
dpgeorge, line 2130: I think we should reserve 0 as a possible valid channel in the future, and instead always use -1 to mean "no channel".

To make it clearer, probably good to define a constant like:
```c
#define CYW43_CHANNEL_NONE (0xffffffff)
```

src/cyw43_ll.c
```diff
@@ -2127,9 +2127,11 @@ int cyw43_ll_wifi_join(cyw43_ll_t *self_in, size_t ssid_len, const uint8_t *ssid
...
+            cyw43_put_le32(buf + 4 + 32 + 20 + 8, 1); // chanspec_num
+            uint16_t chspec = channel | WL_CHANSPEC_BW_20 | WL_CHANSPEC_CTL_SB_NONE | WL_CHANSPEC_BAND_2G;
+            cyw43_put_le16(buf + 4 + 32 + 20 + 12, chspec); // chanspec_list
```
dpgeorge, line 2133: In the `cyw43_write_iovar_n()` call below, should probably have length as `4 + 32 + 20 + 12` when `chanspec_num=0`.

peterharperuk: Would you mind if I keep it at 70 bytes? It matches what the infineon code does - they always send a struct of 70 bytes and just zero the buffer when not setting a channel.

src/cyw43_ll.c
peterharperuk: Yes - will fix

src/cyw43_ll.c
dpgeorge: Yes that's fine, let's keep it as-is.

dpgeorge:
I did some tests with MicroPython and Pico W, and indeed this fixes the case of joining an AP by specifying the BSSID.

PR #58, opened 2023-02-01
cyw43_ll: Support using this driver on stm32 (pybd, portenta).

jimmo:
Minor changes to make this version of the driver behave identically to the version that was previously used in MicroPython for stm32 boards such as GR pybd and Arduino Portenta.

The driver worked as-is, but with two performance regressions fixed by this PR.
- The SDIO interrupt was not being handled correctly.
- The 1DX module (used on both boards) used needs to set revision 17 when using the "worldwide" region.

Also makes it return `EINVAL` rather than `-1` (i.e. `EPERM`) when setting an invalid auth type for connection.

_This work was funded through GitHub Sponsors._

### DIFF
--- a/src/cyw43_ll.c
+++ b/src/cyw43_ll.c
@@ -2061,7 +2061,7 @@ int cyw43_ll_wifi_join(cyw43_ll_t *self_in, size_t ssid_len, const uint8_t *ssid
         wpa_auth = CYW43_WPA_AUTH_PSK;
     } else {
         // Unsupported auth_type (security) value.
-        return -1;
+        return -CYW43_EINVAL;
     }
 
     CYW43_VDEBUG("Setting wsec=0x%lx\n", auth_type & 0xff);
--- a/src/cyw43_ll.c
+++ b/src/cyw43_ll.c
@@ -38,13 +38,13 @@
 #include <inttypes.h>
 
 #include "cyw43_config.h"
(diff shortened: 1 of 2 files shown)


src/cyw43_ll.c
```diff
@@ -44,7 +44,10 @@
 
 #include CYW43_WIFI_NVRAM_INCLUDE_FILE
 
-#define CLEAR_SDIO_INT 0
+#ifndef CYW43_CLEAR_SDIO_INT
+#define CYW43_CLEAR_SDIO_INT 0
+#endif
```
dpgeorge, line 50: please put this in cyw43_config.h

src/cyw43_ll.c
```diff
@@ -1849,7 +1850,7 @@ int cyw43_ll_wifi_on(cyw43_ll_t *self_in, uint32_t country) {
     if (country == CYW43_COUNTRY_WORLDWIDE) {
         // For the 1DX CLM, we need to use revision 17 for the worldwide
         // country. This increases throughput by about 25% on PYBD.
-        country |= (17 << 16)
+        country |= (17 << 16);
```
dpgeorge, line 1853: please squash this fix into the other commit (also the include of cyw43_country.h)

jimmo:
Updated.

dpgeorge:
Thank you!

PR #60, opened 2023-02-06
src: Add changes for SPI Bluetooth support.

peterharperuk:
See CYW43_ENABLE_BLUETOOTH.
New firmware added that supports Wifi and BT.
Firmware now stored as binary data in header files.


firmware/README.md
```diff
@@ -4,18 +4,26 @@ CYW43xx WiFi SoC firmware
...
-Note that the number of dots in the filename is significant, due to the way symbols
-are renamed by the linker when it is included in a compiled application.
+This binary is converted to a header file, e.g. xxd -i 43439A0-7.95.49.00.combined
+The macros `CYW43_WIFI_FW_LEN`, `CYW43_CLM_LEN` specify the unpadded size of the original firmware binaries in bytes.
 
-When updating this firmware, check and update the values of `CYW43_FW_LEN` and
-`CYW43_CLM_LEN`.  These should be the original, unpadded size of the two binaries
-in bytes.  The array size should be the size of the combined file.
+The BT firmware is supplied as a source file containing a static array with ascii hex data. We process this to reduce it in size.
+See CYW43_USE_HEX_BTFW in the SDK to load the original source file. The BT fw is now formatted like this...
```
dpgeorge, line 20: Probable best not to reference the (pico?) SDK here.  Does the format of the BT FW really matter here, ie is any of the following information relevant to this repository?

peterharperuk: I'll remove it. It's only really relevant if you ever get BT firmware from Infineon. I've reprocessed what they gave us (where each byte was encoded by two ascii characters) into something less verbose.

dpgeorge: Ah, I see.  So maybe it is worth having some information about the format used here.  Eg say: "The Bluetooth firmware has the following format: ..."

firmware/README.md
```diff
@@ -4,18 +4,26 @@ CYW43xx WiFi SoC firmware
...
+
+1 byte: number of characters in version string including null terminator
+n bytes: zero terminated version string
+1 byte: number of records
+...
```
dpgeorge, line 25: I guess these dots mean "each record has the following form"?  If so would be good to write that explicitly.

firmware/README.md
```diff
@@ -4,18 +4,26 @@ CYW43xx WiFi SoC firmware
 This directory contains firmware patch blobs that need to be downloaded on to the
 CYW43xx SoC in order for it to function correctly.
 
-The firmware is padded to 512 bytes and then the CLM appended to that, to create the
-combined binary file.
+The Wifi firmware and CLM data is padded to 512 bytes. Then the Wifi firmware, CLM data and BT firmware are all appended together,
+to create the combined binary file.
```
dpgeorge, line 8: The BT firmware is no longer appended, so maybe say: "The WiFi firmware is padded to 512 bytes and then the CLM appended to that, to create the combined WiFi and CLM binary data."

src/cyw43.h
```diff
@@ -619,6 +623,34 @@ static inline uint32_t cyw43_pm_value(uint8_t pm_mode, uint16_t pm2_sleep_ret_ms
  */
 #define CYW43_PERFORMANCE_PM cyw43_pm_value(CYW43_PM2_POWERSAVE_MODE, 20, 1, 1, 1)
 
+#if CYW43_ENABLE_BLUETOOTH
+/*!
+ * \brief Initialise bluetooth
+ *
+ * \return true on success
```
dpgeorge, line 630: returns 0 on success

src/cyw43_ctrl.c
```diff
@@ -65,6 +69,10 @@ cyw43_t cyw43_state;
 void (*cyw43_poll)(void);
 uint32_t cyw43_sleep;
 
+#if CYW43_ENABLE_BLUETOOTH
+void cyw43_hci_process_func(void);
```
dpgeorge, line 73: I think this should be called `cyw43_bluetooth_hci_process()` and declared alongside the other `cyw43_bluetooth_hci_xxx` functions.  At the very least it should go in a header somewhere.

firmware/README.md
peterharperuk: I wrote this - ok? "The WiFi firmware is padded to 512 bytes and then the CLM appended to that to create the combined binary file."

dpgeorge: looks good

dpgeorge:
I've now merged this with the following minor changes (which were force pushed to this PR branch):
- tweaked comments for new hci functions (changed "bt hci" to "Bluetooth HCI layer")
- removed `assert(false)` from `cyw43_ll_sdpcm_poll_device()`, and `assert(ret == 0)` from `cyw43_write_backplane_mem()`
- changed `CYW43_PRINTF` to `CYW43_WARN` in `cyw43_write_backplane_mem()`

Also in a follow-up commit I converted the remaining combined binary blob (for the Murata 1DX) to a header file.

peterharperuk:
Thanks!

peterharperuk:
Yes. It'll be in the pico-sdk for the time being.

PR #64, opened 2023-03-01
firmware: Align firmware blobs.

iabdalkader:
* The firmware blobs need to aligned for loading with the DMA at least on the stm32 port, since the big_const symbol was removed this is no longer controllable by the linker script.
* Fixes firmware loading on Portenta, Giga, Nicla.


peterharperuk:
Seems reasonable. I guess we broke that when moving to header files.

dpgeorge:
It is possible to do this using the linker script, you can explicitly align things using `. = ALIGN(4);` and then place the firmware blobs right after that align.

That said, I think it's cleaner to do this alignment as it's done in this PR, in the C code.

But it should be configurable.

firmware/w43439A0_7_95_49_00_combined.h
```diff
@@ -1,4 +1,4 @@
-static const unsigned char w43439A0_7_95_49_00_combined[] = {
+static const unsigned char w43439A0_7_95_49_00_combined[] __attribute__((aligned(4))) = {
```
dpgeorge, line 1: I suggest making this `CYW43_FIRMWARE_ATTRIBUTE` and then a port can do:
```c
#define CYW43_FIRMWARE_ATTRIBUTE __attribute__((aligned(4)))
```

iabdalkader:
> It is possible to do this using the linker script, you can explicitly align things using `. = ALIGN(4);` and then place the firmware blobs right after that align.

Yes but you will need to know the name of the firmware blob and to change back all the linker script.

> That said, I think it's cleaner to do this alignment as it's done in this PR, in the C code.

I had some doubts about one alignment fit all, but I checked and this works for F7 and H7, that said I will still do the suggested change above.

dpgeorge:
I renamed the config option to `CYW43_RESOURCE_ATTRIBUTE` and applied it also to the NVRAM (which was already aligned, so this makes it consistent with the other blobs).

Thanks for finding this issue!

PR #66, opened 2023-03-11
Add option to allow disabling LWIP

TiejunMS:
I ported NetX running on Pico W and need to disable LWIP. Projects here: https://github.com/TiejunMS/Azure-RTOS-on-Raspberry-Pi-Pico-RP2040/tree/main/demo_netx


peterharperuk:
Why don't you just set CYW43_LWIP=0. What's the purpose of adding another macro?

TiejunMS:
Good suggestion. Thank you!

PR #67, opened 2023-03-11
src/cyw43_ll: Fix missing newline after some WARN and DEBUG messages.

oliver-joos:
I saw this bug when I enabled debug messages in the new cyw43-driver by adding the following 3 lines to cyw43_configport.h:

```
#define CYW43_DEBUG(...) printf(__VA_ARGS__)
#define CYW43_VDEBUG(...) printf(__VA_ARGS__)
#define CYW43_VERBOSE_DEBUG 1
```


dpgeorge:
Thanks for this!  I did notice these missing newlines but didn't get around to fixing them.

PR #70, opened 2023-03-21
Add dependency generation to Makefile

urbasus:
Dependency generation makes sure targets are determined out of date should any included headers be out of date.

Tested by rebuilding across changes to cyw43_configport.h.

### DIFF
--- a/tests/sdio/Makefile
+++ b/tests/sdio/Makefile
@@ -47,3 +47,11 @@ OBJ_DIRS = $(sort $(dir $(OBJ)))
 $(OBJ): | $(OBJ_DIRS)
 $(OBJ_DIRS):
 	$(MKDIR) -p $@
+
+# Dependency generation
+%.o: %.d
+CFLAGS += -MP -MMD
+DEP := $(OBJ:.o=.d)
+$(DEP):
+
+-include $(DEP)


tests/sdio/Makefile
```diff
@@ -47,3 +47,10 @@ OBJ_DIRS = $(sort $(dir $(OBJ)))
 $(OBJ): | $(OBJ_DIRS)
 $(OBJ_DIRS):
 	$(MKDIR) -p $@
+
+# Dependency generation
+CFLAGS += -MP -MMD
+DEP := $(OBJ:.o=.d)
+$(DEP):
+
+-include $(DEP)
```
urbasus, line 56: Include generated headers, if any.

tests/sdio/Makefile
```diff
@@ -47,3 +47,11 @@ OBJ_DIRS = $(sort $(dir $(OBJ)))
 $(OBJ): | $(OBJ_DIRS)
 $(OBJ_DIRS):
 	$(MKDIR) -p $@
+
+# Dependency generation
+%.o: %.d
```
urbasus, line 52: Make sure object is rebuilt if dependency file is out of date.

tests/sdio/Makefile
```diff
@@ -47,3 +47,11 @@ OBJ_DIRS = $(sort $(dir $(OBJ)))
...
+CFLAGS += -MP -MMD
+DEP := $(OBJ:.o=.d)
+$(DEP):
```
urbasus, line 55: Add dummy recipes for dependency files to avoid "no rule to make target" error.

Dependency files are instead created on compiling the object file.

dpgeorge:
Thanks for the contribution.

I actually consciously left this out because I wanted to keep the Makefile as simple as possible.  I like simplicity!

But I guess it doesn't harm to add it, and it's only a few lines.

urbasus:
> Thanks for the contribution.
> 
> I actually consciously left this out because I wanted to keep the Makefile as simple as possible. I like simplicity!
> 
> But I guess it doesn't harm to add it, and it's only a few lines.

I usually leave it out too for the same reason. This time felt like a amusing evening challenge. It took me a couple iterations to get right and terse. If only I could remember the incantation for future projects...

Do as you wish. I trust your judgement.

dpgeorge:
> It took me a couple iterations to get right and terse.

Your solution is very neat.

Merged!

PR #73, opened 2023-03-24
Add include-what-you-use test for public headers

urbasus:
For each public header a test is added to see whether it can be compiled into a pre-compiled header. Success indicates the header includes all the necessary headers.

Headers missing includes have been fixed.

Testing added to phony "test" target.

Dependency generation included.

### DIFF
--- a/src/cyw43_spi.h
+++ b/src/cyw43_spi.h
@@ -34,6 +34,11 @@
 #ifndef CYW43_INCLUDED_CYW43_SPI_H
 #define CYW43_INCLUDED_CYW43_SPI_H
 
+#include "cyw43_internal.h"
+
+#include <stdint.h>
+#include <stddef.h>
+
 // Test register value
 #define TEST_PATTERN 0xFEEDBEADu
 
--- a/tests/sdio/Makefile
+++ b/tests/sdio/Makefile
@@ -48,10 +48,32 @@ $(OBJ): | $(OBJ_DIRS)
 $(OBJ_DIRS):
 	$(MKDIR) -p $@
 
+# "Include-what-you-use"-test compilation for public headers
+PUBLIC_HEADER := \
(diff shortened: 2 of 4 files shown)


src/cyw43_spi.h
```diff
@@ -34,6 +34,11 @@
 #ifndef CYW43_INCLUDED_CYW43_SPI_H
 #define CYW43_INCLUDED_CYW43_SPI_H
 
+#include "cyw43_internal.h"
```
dpgeorge, line 37: That seems strange that a public header needs to include an internal header... need to think about this some more.

tests/sdio/Makefile
```diff
@@ -48,10 +48,32 @@ $(OBJ): | $(OBJ_DIRS)
...
+
+PUBLIC_HEADER_GCH := $(patsubst %,$(BUILD)/%.gch,$(PUBLIC_HEADER))
+
+# Static pattern rule testing header compilation by making a pre-compiled header of each
+$(PUBLIC_HEADER_GCH): $(BUILD)/%.h.gch: %.h
+# Pass -Wno-pedantic to suppress "empty translation unit" errors
+	$(CC) $(CFLAGS) -Wno-pedantic -c $< -o $@
+
+# Add to phony test target trigger
+test: $(PUBLIC_HEADER_GCH)
```
dpgeorge, line 70: Can we move this header test (pre-compiled headers) to a completely separate test?  Eg `tests/headers`/?

src/cyw43_spi.h
urbasus: > That seems strange that a public header needs to include an internal header... need to think about this some more.

Agreed. The supposedly internal cyw43_int_t in the supposedly public API makes for a visibilty mismatch.

I'll handle the other remarks in the meantime.

PR #75, opened 2023-05-02
cyw43_ctrl.c: Use port-defined extint_set and GPIO IT constants.

iabdalkader:
The `extint_set` function and `GPIO_IT_` constants are only defined for the `stm32` port. To improve the portability of the host-wake/OOB mode, those names are replaced with `CYW43_` constants to be defined in ports `cyw43_configport.h`files, when they're used.

NOTE this is a breaking change which will require updating at least the `stm32/cyw43_configport.h` file.


iabdalkader:
Oh I see this change has already been made! closing.

dpgeorge:
Yep!  I also noticed this and fixed it just before the v1.0.0 release of this driver (and corresponding v1.20.0 MP released).

PR #81, opened 2023-06-21
Fix get_stas() bug

heeres:
- cyw43_ll_wifi_ap_get_stas does not perform checks on the return value of cyw43_get_ioctl. On error (e.g. timeout), invalid num_stas values can be used in the following memcpy call, leading to crashes
- cyw43_ll_wifi_ap_get_stas now returns the return code of cyw43_do_ioctl on error
- Perform bounds checks before using memcpy. This requires the  parameter num_stas to contain the size (in MAC addresses) of the supplied buffer.
- Associated functions in cyw43_ctrl.c set num_stas to 0 on error


dpgeorge:
Thanks for the fix!

PR #83, opened 2023-07-04
Guard cyw43_sdio implementation with #if !CYW43_USE_SPI.

esky-software:
This means the file doesn't have to be excluded from the build if CYW43_USE_SPI is defined.


dpgeorge:
Thanks for the patch!

PR #86, opened 2023-08-02
Update cyw43_country.h

engrzamanimran:
Added Pakistan code.


felixdoerre:
Did you see https://github.com/georgerobotics/cyw43-driver/pull/46? The country codes that are defined should actually be the same list that the firmware supports currently. So just adding a new country code will probably not work.

PR #88, opened 2023-09-26
fixing compiler warnings

beechwoods-software:
The lack of casts in these debug messages was resulting in compiler warnings in my enviornment.

### DIFF
--- a/src/cyw43_ll.c
+++ b/src/cyw43_ll.c
@@ -361,7 +361,7 @@ static int cyw43_download_resource(cyw43_int_t *self, uint32_t addr, size_t raw_
     // round up len to simplify download
     size_t len = (raw_len + 255) & ~255;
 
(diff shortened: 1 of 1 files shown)


dpgeorge:
Thanks for the contribution.

Instead of this approach, I'd rather change the format codes to all use `%u` and `%d` and `%x`, ie just use the non-long versions.  That's more suitable to embedded development (smaller strings, no need for the embedded printf to support long args).

Feel free to update this PR to do this, or otherwise I can do it.

beechwoods-software:
> Thanks for the contribution.
> 
> Instead of this approach, I'd rather change the format codes to all use `%u` and `%d` and `%x`, ie just use the non-long versions. That's more suitable to embedded development (smaller strings, no need for the embedded printf to support long args).
> 
> Feel free to update this PR to do this, or otherwise I can do it.

I guess I just did it on a different PR. Probably easiest to just close this one and use the other?

PR #94, opened 2023-10-24
Update firmware for Data Length Extension fix

### LINKED ISSUE #93: Bluetooth Data Length Extensions (DLE) not working properly
We need updated firmware to fix this issue https://github.com/raspberrypi/pico-sdk/issues/1465
The fix improves performance.

peterharperuk:
Bluetooth LE DLE does not work. This updated firmware fixes the problem. See https://github.com/raspberrypi/pico-sdk/issues/1465

Fixes #93


peterharperuk:
I've tested this. Seems to work.

dpgeorge:
Thanks for this.

Does the BT firmware have a version number that we can record somewhere, for reference?

I will do some tests with this new firmware.

peterharperuk:
Hi, it's CYW4343A2_001.003.016.0065.0000_Generic_SDIO_37MHz_wlbga_BU_RPI_dl_signed. I'll add the version to the source file as a comment in future.

Note that we're hoping to get some more fixes soonish if you want to hold off testing until they arrive. Micropython doesn't enable DLE support in btstack - maybe I should push a fix for that? We'll be changing the pico-examples to enable DLE support in the next release.

matsobdev:
Hi! I was playing with this firmware and noticed one thing. There is a vendor specific command: "Set_Transmit_Power" (0xFC26) that apparently changes TX power. With a handle parameter of 0xFE it changes TX power for LE advertisements. But there is an odd behaviour when it doesn't accept valid values, and after cycling through other values it accepts that previous good but discarded at the beginning. Status code 0x00 - success, 0x30 - value out of range. I've made USB serial <-> CYW43439 HCI bridge on Pico W, so it can be connected to AIROC™ Bluetooth® Test and Debug Tool (CYW43439 is supported by this app according to its manual) or just a terminal. Output below:

capture_17032024_212843

Log from ABTaDT: [airocbluetoothtoolHciLog_2024_03_17_21_58_19.txt](https://github.com/georgerobotics/cyw43-driver/files/14628941/airocbluetoothtoolHciLog_2024_03_17_21_58_19.txt)

Other thing with adv (might be a wrong issue, but might be firmware related) is that connectable unidirected adv working good for interval (both min and max) about 100 ms. 150, 200 ms will produce about 800 ms (according to nRF Connect app on the phone). Default value of 0x800 slots = 1280 ms will give scanned interval of 6500 ms highly variable. Non-connectable unidirected adv both high and low values of intervals (min and max the same) acts predictable as scanned with nRF Connect.

peterharperuk:
Sorry @matsobdev I don't really understand the point you're making. Are you saying there's a regression in this firmware?

matsobdev:
Sorry, I was wrong, first is not caused by new firmware (or maybe it is) since old one behaves the same (so maybe still firmware related), it might be ROM issue if it is one. But no, there is no regression. It is stable. The deal is, there is a power table, and values -16, -12, -8, -4, 0, 4, 8, 12 dBm  are right ones. If applied they change RSSI level at receiving end. But still incrementing up from -16 to 8 by one, every value is accepted (0x00 success code), as well as decrementing from 8 to -16 by one. Setting value to 9, 10, 11, 12 changes TX power to the same level (same RSSI), they are all accepted (0x00 - success). But now there is only way to go down one by one, ex. from 10 to 9 (success code), but from 10 to 11 or 12 (error code 0x30 - value out of range). To 'unlock', changing to 8 or lower right one and then going to 11 ot 12 is accepted. I guess it is a bit of an off topic here and if values from -16 to 12 in steps of 4 are legit one, then poking it with other values is not a good idea after all, but still kind of an interesting behaviour. I guess it is the way it is.

Second one with adv intervals manifest also in older firmware, so maybe issue on Pico SDK will be a better place for it. It is not related to BTstack - it happens with or without it.

peterharperuk:
@dpgeorge My earlier optimism that we might get some more fixes to the firmware seems to have been misguided. We might want this DLE fix after all. I will test again and remove the draft label.

dpgeorge:
@peterharperuk do you want to get this fix in?  I can run our standard test suite, but that won't test DLE.

peterharperuk:
I've revived this and retested it. I re-ran the Micropython tests and see no change as far as I can tell. It would be nice to merge it if possible.

shermp:
Hi, is there any chance this could be merged so it can be included in the Pico SDK? Having to patch the SDK driver is getting rather annoying, and the fix does work.

dpgeorge:
Sorry, I didn't realise this was ready to go in now.

I have tested it on a Pico 2 W, running the MicroPython BLE multitests.  It passes.

dpgeorge (approved):

PR #95, opened 2023-11-03
Fix ap shutdown itf_state

drensber:
Without this change, there's really no way to get the CYW43_ITF_AP bit in self->its_state cleared after leaving AP mode.


drensber:
Oops, sorry.... I submitted more changes than I meant to here. I'll try to fix that.

PR #96, opened 2023-11-03
updating itf_state when AP gets shut down

drensber:
Without this change, it's impossible to get self->itf_state's CYW43_ITF_AP bit cleared.


drensber:
I'm having trouble seeing why the formatting check failed. It looks like the new line is aligned with the one above it.

dpgeorge:
It could be because there are tabs instead of spaces?

drensber:
> It could be because there are tabs instead of spaces?

That was it. I didn't even think of that, because I assumed that the GitHub change browser would show stuff like that, but I guess it doesn't (at least in its default configuration). Looks like it's good now.

dpgeorge:
I tested this, bringing up and down the AP multiple times, and it seems to work fine.

PR #98, opened 2023-11-12
Fix arg size in debug message

eightycc:
Debug message argument size and format specifier size do not match.


dpgeorge:
Thanks for the patch, but this is undoing what was done in commit 86a914927cb8a1aa4816dd5ba174ec75f8f75c6a

When do you see an error compiling this bit of code?

eightycc:
Attempting to collect more info, I was unable to reproduce this problem. Sorry for the false alarm.

Issue #106, opened 2023-12-14
AP Mode Cannot Be Restarted After `cyw43_arch_disable_ap_mode`


eightycc:
CircuitPython attempts to restart AP mode (`cyw43_arch_enable_ap_mode`) subsequent to its having been previously started (`cyw43_arch_enable_ap_mode`) and stopped (`cyw43_arch_disable_ap_mode`). All subsequent attempts to restart `cyw43_tcpip_link_status` return `CYW43_LINK_DOWN`.

Tracing indicates that this is due to a race condition between `cyw43_wifi_setup` and asynchronous setting of `NETIF_FLAG_LINK_UP` by `netif_set_link_up` in LWIP.

On the first start of the AP, the CYW43439 takes longer to bring the link up and signal asynchronously than it does on subsequent attempts. For the first AP start `cyw43_wifi_setup()` is able to complete `CYW43_cb_tcpip_init()` before the CYW43439 signals that the link is up, so `NETIF_FLAG_LINK_UP` does not get clobbered. On subsequent starts of the AP, the CYW43439 signals that the link is up almost immediately so `CYW43_cb_tcpip_init()` clears `NETIF_FLAG_LINK_UP` after it has already been set.

See https://github.com/adafruit/circuitpython/issues/8718 for additional details.

dpgeorge:
Fixed by 089a75340e88f6a6d4787e50f9439940af80e11f

PR #107, opened 2023-12-15
Fix race in AP interface set up.

eightycc:
Resolves race condition in #106 by moving the call to `cyw43_wifi_ap_set_up` after the call to `cyw43_cb_tcpip_init` in `cyw43_wifi_set_up`. Since bringing the link up by `cyw43_wifi_ap_set_up` completes asynchronously it is now guaranteed that the setting of `NETIF_FLAG_LINK_UP` will not get clobbered by `cyw43_cb_tcpip_init` if it finishes too early.

This will also resolve https://github.com/adafruit/circuitpython/issues/8718.


dpgeorge:
Thanks for this, it's a good fix.  Sorry it didn't get merged earlier.

PR #108, opened 2023-12-15
Use memmove for operands coincident in memory.

### LINKED ISSUE #99: `memcpy` operands overlap in `spid_buf`
Compiling with `gcc 13.2` at `-O3` has more aggressive in-lining causing detection of illegal `memcpy` operand overlap cases that were not found by earlier versions of `gcc`:

With tag `v1.0.2`
```
In function 'cyw43_send_ioctl',
    inlined from 'cyw43_do_ioctl' at lib/cyw43-driver/src/cyw43_ll.c:1184:15,
    inlined from 'cyw43_set_ioctl_u32' at lib/cyw43-driver/src/cyw43_ll.c:1822:5,
    inlined from 'cyw43_ll_wifi_on' at lib/cyw43-driver/src/cyw43_ll.c:1897:5:
lib/cyw43-driver/src/cyw43_ll.c:772:5: error: 'memcpy' accessing 4 bytes at offsets 68 and 68 overlaps 4 bytes at offset 68 [-Werror=restrict]
  772 |     memcpy(self->spid_buf + SDPCM_HEADER_LEN + 16, buf, len);
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
In this case `memcpy` source and destination coincide at identical offsets inside `spid_buf`.

Same error detected by `gcc 13.2` at several other locations in `cyw43_ll.c`.

Worked around problem by offsetting temporary operand locations in `spid_buf` by + 4. For example:
```c
static void cyw43_set_ioctl_u32(cyw43_int_t *self, uint32_t cmd, uint32_t val, uint32_t iface) {
    uint8_t *buf = &self->spid_buf[SDPCM_HEADER_LEN + 16 + 4];
    cyw43_put_le32(buf, val);
    cyw43_do_ioctl(self, SDPCM_SET, cmd, 4, buf, iface);
}
```

This quiets `gcc` error detection.

eightycc:
Resolves #99 by using `memmove` instead of `memcpy` for operands that are coincident in memory. This pull also resolves compile errors when attempting integration of the latest `cyw43-driver` into CircuitPython 9.0.0.

### DIFF
--- a/src/cyw43_ll.c
+++ b/src/cyw43_ll.c
@@ -746,7 +746,7 @@ static int cyw43_send_ioctl(cyw43_int_t *self, uint32_t kind, uint32_t cmd, size
     header->status = 0;
 
     // copy in payload
-    memcpy(self->spid_buf + SDPCM_HEADER_LEN + 16, buf, len);
+    memmove(self->spid_buf + SDPCM_HEADER_LEN + 16, buf, len);
 
     // do transfer
     CYW43_VDEBUG("Sending cmd %s (%u) len %u flags %u status %u\n", ioctl_cmd_name(header->cmd), header->cmd, header->len, header->flags, header->status);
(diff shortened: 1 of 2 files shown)


dpgeorge:
Thanks for the fix.  I also added a commit on this PR to run CI with `-O3` so this doesn't regress.

PR #119, opened 2024-06-24
WPA3 support

peterharperuk:


B3rn475:
**FYI:** I tested it on my WPA3 only setup and it works like a charm.

Thank you @peterharperuk for implementing the feature this fast.

src/cyw43_ll.c
```diff
@@ -2099,13 +2109,22 @@ int cyw43_ll_wifi_join(cyw43_ll_t *self_in, size_t ssid_len, const uint8_t *ssid
         cyw43_do_ioctl(self, SDPCM_SET, WLC_SET_WSEC_PMK, 68, buf, WWD_STA_INTERFACE); // 68, see wsec_pmk_t
     }
 
+    if (wpa_auth == CYW43_WPA3_AUTH_SAE_PSK) {
+        memset(buf, 0, 128 + 2);
+        cyw43_put_le16(buf, key_len);
+        memcpy(buf + 2, key, key_len);
+        cyw43_delay_ms(2); // Delay required to allow radio firmware to be ready to receive PMK and avoid intermittent failure
+        cyw43_write_iovar_n(self, "sae_password", 2 + 128, buf, WWD_STA_INTERFACE);
```
dpgeorge, line 2117: Where did you get this 2+128 from?  It looks like it should be 2+64:
```c
#define WSEC_MAX_SAE_PASSWORD_LEN      64
typedef struct {
    uint16_t  password_len;             /* octets in key materials  */
    uint8_t   password[WSEC_MAX_SAE_PASSWORD_LEN];    /* maximum key len for SAE passphrase */
} wsec_sae_password_t;
```

src/cyw43_ll.h
```diff
@@ -163,6 +163,7 @@
 // Values used for STA and AP auth settings
 #define CYW43_WPA_AUTH_PSK (0x0004)
 #define CYW43_WPA2_AUTH_PSK (0x0080)
+#define CYW43_WPA3_AUTH_SAE_PSK (0x40000)
```
dpgeorge, line 166: I think these constants should be private to `cyw43_ll.c`.  Maybe we should move them there so they don't confuse users searching for auth settings?

src/cyw43_ll.c
peterharperuk: I guess it's probably ok to leave it as 64 bytes to be consistent with WPA2. I think WPA3 allows up to 128 characters which is where I think I got it from. I'll change this.

src/cyw43_ll.h
peterharperuk: Yes I agree.

src/cyw43_ll.c
peterharperuk: Actually I think I've changed my mind https://github.com/Infineon/wifi-host-driver/blob/master/WiFi_Host_Driver/src/include/whd_wlioctl.h#L748

peterharperuk: I changed my mind because I believe the firmware is expected 128 bytes, e.g. https://github.com/Infineon/wifi-host-driver/blob/master/WiFi_Host_Driver/src/include/whd_wlioctl.h#L752

dpgeorge:
Thanks for adding the password length constants, that's a good improvement.

PR #123, opened 2024-07-18
src/cyw43_bt: Add Bluetooth driver and a config option for the firmware.

iabdalkader:
The Bluetooth driver `cywbt.c` should be moved to this repo first for this path to work. I've made some recent changes to that file that made *most* of it portable, except for a single function that's stm32-specific, but I think it can still be moved. If this is merged I can send a PR to micropython to update the submodule and remove `cywbt.c` from micropython.

See https://github.com/micropython/micropython/pull/11397#issuecomment-1701189783

Note the first commit adds the upstream driver without changes.


src/cyw43_config.h
```diff
@@ -77,6 +77,13 @@
 #define CYW43_WIFI_NVRAM_INCLUDE_FILE "firmware/wifi_nvram_43439.h"
 #endif
 
+// This include should define:
+// - cyw43_btfw_4343A1[]
```
iabdalkader, line 81: Should probably be called something more generic like `cyw43_btfw`.

iabdalkader:
This probably won't be needed.

dpgeorge:
See #123 for an alternative.

PR #130, opened 2025-01-06
Disconnect in response to validation error

edmcman:
Related discussion: https://github.com/raspberrypi/pico-sdk/issues/2153#issuecomment-2572826160

### DIFF
--- a/src/cyw43_ctrl.c
+++ b/src/cyw43_ctrl.c
@@ -413,6 +413,10 @@ void cyw43_cb_process_async_event(void *cb_data, const cyw43_async_event_t *ev) 
             // PSK_SUP failure
             self->wifi_join_state = WIFI_JOIN_STATE_BADAUTH;
         }
+    } else if (ev->event_type == CYW43_EV_ICV_ERROR) {
+        self->pend_rejoin = true;
+        self->pend_rejoin_wpa = false;
+        cyw43_schedule_internal_poll_dispatch(cyw43_poll_func);
     }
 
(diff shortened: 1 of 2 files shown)


edmcman:
This is a very simple way to handle this event, which worked well on my buggy AP.  But there might be better, more sophisticated ways of handling this event, such as keeping track of how long it's been since the last successfully decrypted packet, and only disconnecting if it has been a while.

peterharperuk:
Looks good to me.

dpgeorge:
Thanks for the PR, this looks good to me, but I will run some tests first to make sure it's all OK.

edmcman:
I don't think I introduced any of the code formatting issues.  Do you want me to run the code formatter and push the result to the PR?

dpgeorge:
> I don't think I introduced any of the code formatting issues. Do you want me to run the code formatter and push the result to the PR?

Ah, I think this is because the GitHub Action's ubuntu-latest image updated recently and it must have changed the version of uncrustify... the fix (for now) would be to use ubuntu-22.04 for the workflow.  See #131 for that.

dpgeorge (approved):
I tested this on PYBD_SF6 and RPI_PICO_W with MicroPython, and there are no regressions that I could see.

I could not test the new event but it looks correct, and your testing shows that it works.

dpgeorge:
Rebased and merged in c1075d4bc440422cf2b2fd12c64a1f53f77660ee

PR #133, opened 2025-01-23
Add name for CYW43_EV_ICV_ERROR to event name table

edmcman:


edmcman:
Minor detail that I forgot in the last PR handling ICV_ERROR. Sorry!

dpgeorge:
Actually, I thought it was reasonable (and on purpose) to leave this out.  It's a rare event and adding strings to this table increases the size of the driver, which should try to stay relatively small for embedded targets.

If you think it's useful to have, we can add it.  But I'm also happy to leave it out.

edmcman:
Ah, I had not considered the size factor.  It would have been nice to have it set for me when I was originally troubleshooting the problem, but since it seems to be a rare problem that is now hopefully fixed anyway, I'd say let's leave it out.  There should be a decent amount of google hits on the string now anyway from all my posts.

PR #135, opened 2025-02-20
src/cyw43_spi: Add support for gSPI transport.

dpgeorge:
This adds support for (general) SPI as a transport backend to communicate with the cyw43xx chip.  To use this transport, a port must provide the `cyw43_spi_xxx` functions defined in `cyw43_spi.h`, and enable the configuration option `CYW43_USE_SPI`.  The interrupt pin is configured as `CYW43_PIN_WL_IRQ`.

This feature was tested on a board using a Murata 1YN connected in gSPI mode.

This PR was also tested on a Pico 2 W and PYBD-SF2 to check for any regressions.  None were found.

### DIFF
--- a/src/cyw43_ctrl.c
+++ b/src/cyw43_ctrl.c
@@ -45,10 +45,12 @@
 #include "cyw43_btbus.h"
 #endif
 
+#if !CYW43_USE_SPI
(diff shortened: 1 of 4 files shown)


dpgeorge:
@iabdalkader FYI

PR #137, opened 2025-02-27
Add Bluetooth UART HCI backend driver

dpgeorge:
This allows using a UART to communicate with the cyw43's Bluetooth subsystem.  It provides basic functions to init/deinit the controller, and control its sleep state.  Once its init'd, the BT stack can communicate directly over the UART to the cyw43 BT controller.


iabdalkader:
Are you sure about this change? Wouldn't this break other drivers and all ports? Why not keep the old `mp_bluetooth_hci` API? The driver could just provide the implementation in `cyw43_bthci.c` as it did, similar to https://github.com/micropython/micropython/blob/master/drivers/esp-hosted/esp_hosted_bthci.c

src/cyw43_btuart.c
```diff
@@ -0,0 +1,297 @@
+/*
```
iabdalkader, line 1: I think this should be called `cyw43_bthci.c` similar to https://github.com/micropython/micropython/blob/master/drivers/esp-hosted/esp_hosted_bthci.c.

dpgeorge:
> Why not keep the old `mp_bluetooth_hci` API?

Because `mp` here stands for MicroPython, and this driver is supposed to be an independent component, with its own naming scheme.

I understand it introduces the need for a shim layer, but we really can't have any MicroPython names in this repo.

src/cyw43_btuart.c
dpgeorge: We need to distinguish this file from a BT HCI driver that uses a different transport, like the SPI/SDIO interface.  Hence the need to put "uart" in the filename.

How about `cyw43_bthci_uart.c`?

iabdalkader: Yes, `xx_bthci_uart.c` is good. The other drivers above should be renamed to match this.

iabdalkader:
> I understand it introduces the need for a shim layer

A shim layer would minimize changes to `mpbthciport.c` files and other drivers and avoid breaking the somewhat existing patterns. I assume it will implement the weak functions to call CYW43 functions? Something like this for example?
```C
int mp_bluetooth_hci_controller_init(void) {
    return cyw43_bluetooth_controller_init();
}
```

src/cyw43_btuart.c
iabdalkader: I've renamed them here to match the cyw43: https://github.com/micropython/micropython/pull/16833.

iabdalkader:
I'd also suggest moving this submodule from `lib/cyw43-driver` to `drivers/cyw43` with the other WiFi drivers. Perhaps in the future they all could be moved to `drivers/wifi/`

src/cyw43_btuart.c
dpgeorge: I've renamed this to `cyw43_bthci_uart.c` in this PR.

dpgeorge:
> A shim layer would minimize changes to `mpbthciport.c` files and other drivers and avoid breaking the somewhat existing patterns.

You could either implement the shim layer as you suggest, like:
```c
int mp_bluetooth_hci_controller_init(void) {
    return cyw43_bluetooth_controller_init();
}
```

Or more simply, in `cyw43_configport.h`, add defines to rename the cyw43 functions:
```c
#define cyw43_bluetooth_controller_init        mp_bluetooth_hci_controller_init
#define cyw43_bluetooth_controller_deinit      mp_bluetooth_hci_controller_deinit
#define cyw43_bluetooth_controller_sleep_maybe mp_bluetooth_hci_controller_sleep_maybe
#define cyw43_bluetooth_controller_woken       mp_bluetooth_hci_controller_woken
#define cyw43_bluetooth_controller_wakeup      mp_bluetooth_hci_controller_wakeup
```
I've tested both schemes and they both work.  The latter is zero overhead.

iabdalkader:
> I've tested both schemes and they both work. The latter is zero overhead.

I like the second option more too.

dpgeorge:
This PR has now been tested in the following scenarios:
- MicroPython stm32 port, PYBD_SF6 board: no changes required, can still use the old `drivers/cyw43/cywbt.c` code
- MicroPython rp2 port, RPI_PICO_W board: same as above, no changes required
- MicroPython stm32 port, PYBD_SF6, with changes to switch to the new HCI UART code here: needs an implementation of `mp_bluetooth_hci_uart_readchar_blocking()` and a handful of configuration options in `cyw43_configport.h`

iabdalkader:
I tested this with mimxrt, but it's not working. I get:

```
>>> [CYW43] cyw43_bluetooth_download_firmware: unexpected response 01
```

Note `CYW43_PIN_BT_CTS` was optional, I had it commented out for a reason, but I enabled it since it's now required. Maybe it's related?


I made the following changes:

```diff
diff --git a/ports/mimxrt/cyw43_configport.h b/ports/mimxrt/cyw43_configport.h
index 619db24b1..48c6ec983 100644
--- a/ports/mimxrt/cyw43_configport.h
+++ b/ports/mimxrt/cyw43_configport.h
@@ -34,10 +34,12 @@
 #include "extmod/modnetwork.h"
 #include "pendsv.h"
 #include "sdio.h"
+#include "extmod/mpbthci.h"
 
 #define CYW43_USE_SPI                   (0)
 #define CYW43_LWIP                      (1)
 #define CYW43_USE_STATS                 (0)
+#define CYW43_BT_USE_UART               (1)
 
 #ifndef CYW43_CHIPSET_FIRMWARE_INCLUDE_FILE
 #define CYW43_CHIPSET_FIRMWARE_INCLUDE_FILE "lib/cyw43-driver/firmware/w4343WA1_7_45_98_50_combined.h"
@@ -47,6 +49,10 @@
 #define CYW43_WIFI_NVRAM_INCLUDE_FILE   "lib/cyw43-driver/firmware/wifi_nvram_1dx.h"
 #endif
 
+#ifndef CYW43_BT_FIRMWARE_INCLUDE_FILE
+#define CYW43_BT_FIRMWARE_INCLUDE_FILE  "lib/cyw43-driver/firmware/cyw43_btfw_4343A1.h"
+#endif
+
 #define CYW43_IOCTL_TIMEOUT_US          (1000000)
 #define CYW43_SLEEP_MAX                 (50)
 #define CYW43_NETUTILS                  (1)
@@ -75,6 +81,7 @@
 #define CYW43_HAL_PIN_PULL_DOWN         MP_HAL_PIN_PULL_DOWN
 
 #define CYW43_HAL_MAC_WLAN0             MP_HAL_MAC_WLAN0
+#define CYW43_HAL_MAC_BDADDR            MP_HAL_MAC_BDADDR
 
 #define cyw43_hal_ticks_us              mp_hal_ticks_us
 #define cyw43_hal_ticks_ms              mp_hal_ticks_ms
@@ -88,9 +95,21 @@
 #define cyw43_hal_get_mac_ascii         mp_hal_get_mac_ascii
 #define cyw43_hal_generate_laa_mac      mp_hal_generate_laa_mac
 
+extern uint8_t mp_bluetooth_hci_cmd_buf[4 + 256];
+#define cyw43_hal_uart_cmd_buf          mp_bluetooth_hci_cmd_buf
+#define cyw43_hal_uart_set_baudrate     mp_bluetooth_hci_uart_set_baudrate
+#define cyw43_hal_uart_write            mp_bluetooth_hci_uart_write
+#define cyw43_hal_uart_readchar         mp_bluetooth_hci_uart_readchar
+
 #define cyw43_delay_us                  mp_hal_delay_us
 #define cyw43_delay_ms                  mp_hal_delay_ms
 
+#define cyw43_bluetooth_controller_init        mp_bluetooth_hci_controller_init
+#define cyw43_bluetooth_controller_deinit      mp_bluetooth_hci_controller_deinit
+#define cyw43_bluetooth_controller_sleep_maybe mp_bluetooth_hci_controller_sleep_maybe
+#define cyw43_bluetooth_controller_woken       mp_bluetooth_hci_controller_woken
+#define cyw43_bluetooth_controller_wakeup      mp_bluetooth_hci_controller_wakeup
+
 #define CYW43_PIN_WL_REG_ON             MICROPY_HW_WL_REG_ON
 #define CYW43_PIN_WL_HOST_WAKE          MICROPY_HW_WL_HOST_WAKE
 #define CYW43_PIN_WL_SDIO_1             MICROPY_HW_SDIO_D1
```

iabdalkader:
I just needed a blocking read function:
```C
int mp_bluetooth_hci_uart_readchar_blocking(void) {
    int c;
    while ((c = mp_bluetooth_hci_uart_readchar()) == -1) {
        mp_event_wait_indefinite();
    }
    return c;
}
```
It's not clear if I can make the old read function blocking or not.

Unrelated, but what's the point of `cyw43_hal_uart_cmd_buf`? It seems to be me it can just be a stack buffer, no?

dpgeorge:
So, is it working now on mimxrt?  Did you need to enable the CTS pin?  I suspect that's necessary.

> Why not implement the blocking read in `cyw43_bthci_uart.c`? This way all ports can share the same logic, and won't have to add a new function to `mpbthciport.c`.

I was thinking that each port needs different behaviour for the blocking UART read, like a different polling function when waiting for incoming chars, depending on how that port does events/irqs.  So it's easier to make the port implement the blocking read, rather than try to make a generic and configurable one here.

iabdalkader:
> So, is it working now on mimxrt? Did you need to enable the CTS pin? I suspect that's necessary.

Yes, it's working. I only tried a simple aioble LED blinky though. CTS must be defined, otherwise it won't build. Unless you're sure that all chips support reading the pin when it's configured as AF, and that CTS will always be connected,  you might want to consider making it optional like it used to be.

> I was thinking that each port needs different behaviour for the blocking UART read

I think most of them will just have the same code. Could we provide a config option like `cyw43_hal_idle`? Or make `cyw43_hal_uart_readchar_blocking` optional? If it's not defined, default blocking behavior used. For example:
```C
int cyw43_uart_readchar() {
    #ifdef cyw43_hal_uart_readchar_blocking
    return cyw43_hal_uart_readchar_blocking();
    #else
    int c;
    uint32_t ticks = ticks_ms();
    while ((c = mp_bluetooth_hci_uart_readchar()) == -1) {        
        if (ticks> timeout) {
           return -1;
         }
         mp_event_wait_indefinite();
    }
    return c;
#endif
}
```

> but what's the point of `cyw43_hal_uart_cmd_buf`? It seems to be me it can just be a stack buffer, no?

Is this really needed?

src/cyw43_bthci_uart.c
```diff
@@ -0,0 +1,297 @@
...
+        // Construct HCI header.
+        uint8_t *buf = cyw43_hal_uart_cmd_buf;
+        memcpy(buf + 1, firmware, 3);
+        firmware += 3;
+        buf[0] = 1;
+        last_packet = buf[1] == 0x4e;
+        uint8_t len = buf[3];
+
+        // Copy in payload.
+        memcpy(buf + 4, firmware, len);
```
iabdalkader, line 153: A sanity check here might be a good idea. For example: `assert(len <= sizeof(cmd_buf))`

src/cyw43_bthci_uart.c
```diff
@@ -0,0 +1,297 @@
...
+    bool last_packet = false;
+    while (!last_packet && firmware < firmware_max) {
+        // Validate firmware packet.
+        if (firmware[1] != 0xfc) {
+            CYW43_WARN("cyw43_bluetooth_download_firmware: unexpected packet %02x\n", firmware[1]);
+            return -CYW43_EIO;
+        }
+
+        // Construct HCI header.
+        uint8_t *buf = cyw43_hal_uart_cmd_buf;
```
iabdalkader, line 145: This buffer can be replaced with a stack buffer `uint8_t cmd_buf[256+4]`. One less config option to worry about.

src/cyw43_bthci_uart.c
dpgeorge: I've refactored this and added an assert.

src/cyw43_bthci_uart.c
dpgeorge: OK, you're right, this buffer could be fully stack allocated.

I've reworked this now and made it so that you can provide an optional static buffer.  If that's not defined then this driver automatically stack allocates it.

dpgeorge:
> CTS must be defined, otherwise it won't build. Unless you're sure that all chips support reading the pin when it's configured as AF, and that CTS will always be connected, you might want to consider making it optional like it used to be.

I'm pretty sure CTS is required, because there are very long delays (70ms eg at initial start up, and then longer than that after downloading the patch firmware), and the only way to know when the chip is ready is to wait for the CTS line to go low.

I don't know of any code that uses the old driver without CTS, so I think we need to make it mandatory.

> I think most of them will just have the same code. Could we provide a config option like `cyw43_hal_idle`? Or make `cyw43_hal_uart_readchar_blocking` optional? If it's not defined, default blocking behavior used.

OK, I've now done something like that, so you can either provide the blocking or non-blocking version.

iabdalkader:
> I'm pretty sure CTS is required, because there are very long delays (70ms eg at initial start up
> I don't know of any code that uses the old driver without CTS, so I think we need to make it mandatory.

FWIW I use the old HCI UART driver on mimxrt port without CTS (we have a board with RT1062 + CYW43), and it works fine. Note that the old driver already had large delays when CTS was not defined and `wait_cts` was only enabled for stm32 port (as it needed to configure AF).

I enabled CTS when I tested this PR and it works fine, so either way is fine with me.

src/cyw43_bthci_uart.c
iabdalkader: Note this buffer is also used in `extmod/nimble/hal/hal_uart.c` in a very similar way. Replacing that one too with a stack buffer could save data memory that was unrecoverable before.
https://github.com/micropython/micropython/blob/master/extmod/nimble/hal/hal_uart.c#L73

iabdalkader:
Looks good. I can retest on mimxrt when it's merged. I can also send a PR to switch mimxrt when the submodule is updated.

src/cyw43_bthci_uart.c
dpgeorge: Yes, we can fix that eventually.

dpgeorge:
Renamed `CYW43_PIN_WL_RFSW_VDD` to `CYW43_PIN_RFSW_VDD` because it's now part of the BT driver (so not just for WL).  A backwards-compatible macro is provided for this so no user code needs to be changed.

Also added some documentation about the pins that can be defined.

Retested with the same method described in https://github.com/georgerobotics/cyw43-driver/pull/137#issuecomment-2690523795

src/cyw43_bthci_uart.c
```diff
@@ -0,0 +1,326 @@
...
+    buf[0] = val;
+    buf[1] = val >> 8;
+}
+
+static inline void cyw43_bluetooth_put_le32(uint8_t *buf, uint32_t val) {
+    buf[0] = val;
+    buf[1] = val >> 8;
+    buf[2] = val >> 16;
+    buf[3] = val >> 24;
+}
```
iabdalkader, line 69: If `CYW43_BT_UART_BAUDRATE_DOWNLOAD_FIRMWARE` and `CYW43_BT_UART_BAUDRATE_ACTIVE_USE` are not defined, I think these static functions will be unused and issue warnings.

dpgeorge: > maybe not, because they're inline

Right, because they are `static inline` they don't need to be used, and you won't get a warning.

> Also, this is probably an overkill, but if the host byte order can change then you might want to add the following, and handle big endian:

Actually, the reason to write these endian functions explicitly like this is so they are portable across both little and big endian archs.  The point is that the HCI command to change the baudrate is specified to take parameters in little endian order, so that's how the packet is created.

dpgeorge:
I've added a commit here which adds a test for the new BT HCI UART code.  It's a very simple test but ensures that it compiles with a minimal configuration, and runs in a basic capacity.

As part of this I found an issue enabling the BT option, so I renamed `CYW43_BT_USE_UART` to `CYW43_ENABLE_BLUETOOTH_OVER_UART`.  And you enable this and not `CYW43_ENABLE_BLUETOOTH` (which is for the SPI/SDIO transport only).

iabdalkader:
Tested again, still working fine. The changes to mimxrt are here: https://github.com/micropython/micropython/pull/16848

src/cyw43_bthci_uart.c
dpgeorge: clang does indeed give a warning about `static inline` functions not being used.  So I've added the #if guard.

iabdalkader: I've had similar issues with clang, but `inline` is merely a suggestion, isn't it? So it wouldn't know if it's called or not until later stages I guess.

PR #151, opened 2026-06-15
src/cyw43_ctrl: Add roaming and interference-mitigation config via ioctl helpers

mitchellcairns:
Follow-up to the discussion on #146. This reworks the WiFi roaming (and
interference-mitigation) configuration to be implemented entirely in terms of
`cyw43_ioctl()`, per @dpgeorge's feedback, instead of adding dedicated
low-level `cyw43_ll_*` functions for each setting.

The goal is to expose the handful of roaming/interference ioctls that are
critical for tuning link stability, while keeping the embedded footprint
minimal: only two small generic helpers cost binary space; everything
feature-specific is `static inline` (or a tiny iovar wrapper) and is only
instantiated when actually used.

## Background / motivation

Roaming is enabled by default with no way to configure it. When a STA's RSSI
drops below the (hidden) roam trigger, the firmware starts background roam
scans, which causes dropped packets and severe throughput degradation on a
point-to-point link. See:

- georgerobotics/cyw43-driver#111
- raspberrypi/pico-sdk#2835

Disabling roaming (or tuning the trigger/delta/scan-period) fully resolves
these issues, but until now there was no API and no visibility
that this was even the cause.

## Implemented

New generic helpers (the only functions that add code size):

- `cyw43_ioctl_set_u32(self, cmd, val, iface)`
- `cyw43_ioctl_get_u32(self, cmd, *out_val, iface)`

  Thin wrappers around `cyw43_ioctl()` for the common case of a 4-byte ioctl
  value. Argument order matches `cyw43_ioctl()` (`iface` last).

New `CYW43_IOCTL_xxx` constants (encoded as `(WLC_cmd << 1) | is_set`, matching
the existing `CYW43_IOCTL_*` defines):

- `GET`/`SET_ROAM_TRIGGER`, `GET`/`SET_ROAM_DELTA`, `GET`/`SET_ROAM_SCAN_PERIOD`
- `GET`/`SET_INTERFERENCE_MODE`


peterharperuk:
Have you done much testing with this? I'm wondering what values for the roaming and interference mode you've found make a difference?

mitchellcairns:
> Have you done much testing with this? I'm wondering what values for the roaming and interference mode you've found make a difference?

In my configurations I've disabled both roaming/interference as the roaming mode only serves to cause data rate issues. The default value is rather aggressive.

In my other testing, using some debug logging to track the signal level in dBm, the roaming value set aligns with what is set using the helpers. 

Note that this change does not alter the default configuration set, it only exposes these helpers so that developers can utilize them with built in, documented options.

mihneaco:
Had this exact issue on a RC car + controller pair — 2 Pico2W (RP2350),
one AP + one STA, talking via UDP. ~10 s after radio-up, usable range
collapsed from >6m to <1m. The link still worked relatively fine at close range with random latency spikes, and a
full deinit/init of the wifi stack reset it for another ~10 s. Spent one week pulling
my hair suspecting EMI from the motors or other hardware related issues before finding this.

Can confirm `cyw43_wifi_set_interference_mode(&cyw43_state, CYW43_IFMODE_NONE)`
from the commit attached to this PR completely fixes the range collapse. Checked with both `cyw43_wifi_set_roam_enabled(false)` and `cyw43_wifi_set_interference_mode(CYW43_IFMODE_NONE)` and `cyw43_wifi_set_interference_mode` alone was enough for my range issue but `cyw43_wifi_set_roam_enabled(false)` probably helps in the long run - will have to test.

For anyone using pico-sdk: `cyw43_wifi_set_interference_mode` and `cyw43_wifi_set_roam_enabled` took effect only after `cyw43_arch_enable_ap_mode()` / `cyw43_arch_enable_sta_mode()`. Calling them immediately after
`cyw43_arch_init()` didn't seem to apply them, most likely values get overwritten during AP/STA bringup.

mitchellcairns:
@peterharperuk Thanks for the review feedback earlier. I wanted to circle back on this since it's been a couple months. 

I'm confident the implementation addresses the real issues users are hitting (see #111 and pico-sdk#2835). Let me know if there's anything blocking merge or if you need any changes.

mitchellcairns:
Hi @dpgeorge, I hope you are doing well. I wanted to check in on this PR.

I answered @peterharperuk's question on Jun 22 about testing and observed values. The change still does not alter the default roaming behavior. It only exposes the ioctl helpers so developers can tune roaming and interference settings when they need to.

This follows the approach you suggested on PR #146: use cyw43_ioctl() directly instead of adding new low-level cyw43_ll_* functions. I kept the footprint small. Only the two generic u32 helpers add binary size.

Let me know if you would like any changes, more testing data, or a different approach. I am glad to update the PR.

Thank you for your time.

src/cyw43.h
```diff
@@ -104,6 +104,20 @@
 #define CYW43_LINK_BADAUTH      (-3)    ///< Authenticatation failure
 //!\}
 
+/*!
+ * \name Interference mitigation mode
+ * \anchor CYW43_IFMODE_
+ * \see cyw43_wifi_set_interference_mode() to set the mitigation mode
+ * \see cyw43_wifi_get_interference_mode() to get the mitigation mode
+ */
```
dpgeorge, line 113: Do you know what the default is?  Might be worth mentioning that here.

src/cyw43.h
```diff
@@ -104,6 +104,20 @@
...
+//!\{
+#define CYW43_IFMODE_NONE         (0)     ///< None/Disabled
+#define CYW43_IFMODE_NONWLAN      (1)     ///< Non-WLAN
+#define CYW43_IFMODE_WLANMANUAL   (2)     ///< Adjacent channel interference (ACI) mode
+#define CYW43_IFMODE_AUTO         (3)     ///< Automatic as determined by driver (ACI)
+#define CYW43_IFMODE_AUTONOISE    (4)     ///< Automatic as determined by driver with noise reduction (ACI)
```
dpgeorge: "IF" can sometimes mean "interface", so maybe make these names a bit longer, eg `CYW43_INTERFERE_NONE`, `CYW43_INTERFERE_NON_WLAN` etc.

Also, I suggest putting underscore between words to make it a bit more readable, eg `NON_WLAN`, `WLAN_MANUAL`, `AUTO_NOISE`.

src/cyw43.h
```diff
@@ -360,6 +431,109 @@ int cyw43_wifi_join(cyw43_t *self, size_t ssid_len, const uint8_t *ssid, size_t
...
+ *
+ * \param self          The driver state object, always \c &cyw43_state
+ * \param trigger_dbm   RSSI threshold in dBm below which roam scanning begins.
+ *                      Typical value: -75. Must be negative.
+ * \param candidate_delta_db  Minimum RSSI improvement in dB a candidate AP
+ *                      must offer over the current AP to trigger a roam.
+ *                      Typical value: 10.
+ * \param scan_period_ms How often in milliseconds the firmware scans for better
+ *                      APs while below the trigger threshold.
+ * \return 0 on success, negative error code on failure
```
dpgeorge, line 490: Do you know the defaults for these?

src/cyw43.h
```diff
@@ -360,6 +431,109 @@ int cyw43_wifi_join(cyw43_t *self, size_t ssid_len, const uint8_t *ssid, size_t
...
+ * \param self                The driver state object, always \c &cyw43_state
+ * \param trigger_dbm         Output: RSSI threshold in dBm below which roam
+ *                            scanning begins. Will be a negative value.
+ * \param candidate_delta_db  Output: Minimum RSSI improvement in dB a candidate
+ *                            AP must offer over the current AP to trigger a roam.
+ * \param scan_period_ms      Output: How often in milliseconds the firmware scans
+ *                            for better APs while below the trigger threshold.
+ * \return 0 on success, negative error code on failure
+ */
+static inline int cyw43_wifi_get_roam_params(cyw43_t *self, int *trigger_dbm, int *candidate_delta_db, int *scan_period_ms) {
```
dpgeorge: Does it make sense to type these params as `int32_t`?  Then they could be passed directly though to `cyw43_ioctl_get_u32()`.

src/cyw43_ctrl.c
```diff
@@ -455,6 +455,22 @@ int cyw43_ioctl(cyw43_t *self, uint32_t cmd, size_t len, uint8_t *buf, uint32_t
...
+    // The wire format is little-endian; RP2 hosts are little-endian so the
+    // value can be passed straight through (matching cyw43_wifi_get_rssi()).
+    uint32_t buf = val;
+    return cyw43_ioctl(self, cmd, sizeof(buf), (uint8_t *)&buf, iface);
+}
+
+int cyw43_ioctl_get_u32(cyw43_t *self, uint32_t cmd, uint32_t *out_val, uint32_t iface) {
+    uint32_t buf = 0;
+    int ret = cyw43_ioctl(self, cmd, sizeof(buf), (uint8_t *)&buf, iface);
+    if (ret == 0 && out_val != NULL) {
```
dpgeorge: Does it make sense to allow `out_val == NULL`?  What reason would you have for calling this function but not caring about the return value?

Could just pass `out_val` directly to `cyw43_ioctl()` if it was assumed non-null.

src/cyw43_ctrl.c
```diff
@@ -455,6 +455,22 @@ int cyw43_ioctl(cyw43_t *self, uint32_t cmd, size_t len, uint8_t *buf, uint32_t
     return ret;
 }
 
+int cyw43_ioctl_set_u32(cyw43_t *self, uint32_t cmd, uint32_t val, uint32_t iface) {
+    // The wire format is little-endian; RP2 hosts are little-endian so the
```
dpgeorge: This code runs on more than RP2xxx MCUs.... it tries to be endianness-clean (see eg `cyw43_put_le32()` for instance) but because probably everything it's used on is little endian it may in some places assume that.

I'd say you can just remove this comment.

dpgeorge (commented):
Thanks @mitchellcairns for reworking this feature.  It's looking good, just a few technical review points to cover.

mitchellcairns:
Thanks for the review. All points are addressed in 25e470e.

I flashed a Pico W and read the defaults back with the new getters. Immediately after cyw43_arch_init():

roam trigger:      -75 dBm
roam delta:        20 dB
roam scan period:  10
roam_off:          0, so roaming is on
interference mode: 4, CYW43_INTERFERE_AUTO_NOISE
The same values come back after cyw43_arch_enable_sta_mode(), so STA bringup leaves them alone. Both sets are in the doxygen now.

Mode 4 is the highest of the five interference settings. That fits what @mihneaco reported above, where CYW43_INTERFERE_NONE on its own restored his range.

The constants are now CYW43_INTERFERE_NONE, NON_WLAN, WLAN_MANUAL, AUTO and AUTO_NOISE.

cyw43_ioctl_get_u32() takes a non-null out_val and hands it straight to cyw43_ioctl(). That drops the local and the copy, and matches how cyw43_wifi_get_rssi() already works.

The roam params are int32_t, so the getters pass the caller's pointers into cyw43_ioctl_get_u32() with no temporaries.

One change beyond your list. The third roam parameter was named scan_period_ms, and a default of 10 makes me think it may not be milliseconds. The firmware does no range checking on it either: every value I sent from 0 to 2147483647 was accepted and read back unchanged, so its behavior pins nothing down. I renamed it to scan_period with no unit in the name and recorded the default of 10.

Round trip on hardware: -80, 15 and 20, roaming off, interference NONE all read back exactly.

src/cyw43_ctrl.c
```diff
@@ -461,6 +455,15 @@ int cyw43_ioctl(cyw43_t *self, uint32_t cmd, size_t len, uint8_t *buf, uint32_t
     return ret;
 }
 
+int cyw43_ioctl_set_u32(cyw43_t *self, uint32_t cmd, uint32_t val, uint32_t iface) {
+    uint32_t buf = val;
+    return cyw43_ioctl(self, cmd, sizeof(buf), (uint8_t *)&buf, iface);
```
dpgeorge: How about just taking the address of `val` directly?

dpgeorge:
Thanks for updating.

I was wondering it we should change the defaults for some of these parameters, but actually they look pretty reasonable.  A roam trigger of -75dBm is already quite low.  And it's probably best to leave the interference value at the AUTO_NOISE value, that seems to be the most sophisticated option.

So maybe leaving them as they are is best, and then good documentation/examples could show how to tune them for specific needs; such docs should go downstream, eg in pico-sdk and MicroPython.

dpgeorge:
To make CI happy, please `git am` this file: 
[0001-pyproject.toml-Add-pyproject.toml-with-a-skip-list-f.patch](https://github.com/user-attachments/files/31764071/0001-pyproject.toml-Add-pyproject.toml-with-a-skip-list-f.patch)

mitchellcairns:
Both done in 8179a4d, cheers

dpgeorge (approved):
Thanks!

Issue #141, opened 2025-11-20
auth_mode in cyw43_ev_scan_result_t is wrong


peterharperuk:
The auth_mode field in cyw43_ev_scan_result_t isn't right. (Did I add the doxygen?)

```
/*!
 * \brief Structure to return wifi scan results
 */
//!\{
typedef struct _cyw43_ev_scan_result_t {
    uint32_t _0[5];
    uint8_t bssid[6];   ///< access point mac address
    uint16_t _1[2];
    uint8_t ssid_len;   ///< length of wlan access point name
    uint8_t ssid[32];   ///< wlan access point name
    uint32_t _2[5];
    uint16_t channel;   ///< wifi channel
    uint16_t _3;
    uint8_t auth_mode;  ///< wifi auth mode \ref CYW43_AUTH_
    int16_t rssi;       ///< signal strength
} cyw43_ev_scan_result_t;
//!\}

```

It's only 8 bits when it needs to be 32 bits. I think it's actually the dtim period if you look where it's pulled from the frame.
I have some code that extracts this value from the RSN information element if you would like this to be fixed? I think it's useful to know if an AP is WPA2 or WPA3?

broughtong:
Just got caught out by this too!

Thought I could take the scan results auth_mode field and then feed it back in to attempt to make a connection...nope!

My scan result can back as '5', lucky I found this thread as even with a bit of digging + docs I really couldn't see how it would map to 0x00200002, 0x00400004, or 0x00400006

dpgeorge:
Fixed by 8a6dcf57539aac3d7da390b2e09335f025cf3af9



> My scan result can back as '5'

That now corresponds to CYW43_AUTH_FLAG_WEP_ENABLED | CYW43_AUTH_FLAG_AES_ENABLED, and probably at least another bit like CYW43_AUTH_FLAG_WPA3_SECURITY which is now exposed by this driver in scan results.

PR #152, opened 2026-06-24
src/cyw43_ll: parse RSN/WPA IEs in wifi scans results

r-c-n:
Add a function (cyw43_ll_wifi_parse_ie_auth) to parse the vendor-specific WPA IEs and RSN IEs received during wifi scans.

Define a more complete set of security and authentication flags and redefine the CYW43_AUTH_ macros in terms of these new flags.

Return the AP security and authentication capabilities as a 32 bit field in cyw43_ev_scan_result_t.auth_mode (previously an 8 bit field). Note that this change may break compatibility with drivers that use the scan feature.

### DIFF
--- a/src/cyw43_ll.h
+++ b/src/cyw43_ll.h
@@ -161,18 +161,40 @@
 #define CYW43_REASON_SUP_DEAUTH             (14)    // received FC_DEAUTH
 #define CYW43_REASON_SUP_WPA_PSK_TMO        (15)    // WPA PSK 4-way handshake timeout
 
+// Security and authentication flags
+#define CYW43_AUTH_FLAG_WEP_ENABLED              0x0001
+#define CYW43_AUTH_FLAG_TKIP_ENABLED             0x0002
+#define CYW43_AUTH_FLAG_AES_ENABLED              0x0004
+#define CYW43_AUTH_FLAG_SHARED_ENABLED       0x00008000
+#define CYW43_AUTH_FLAG_WPA_SECURITY         0x00200000
+#define CYW43_AUTH_FLAG_WPA2_SECURITY        0x00400000
+#define CYW43_AUTH_FLAG_WPA2_SHA256_SECURITY 0x00800000
+#define CYW43_AUTH_FLAG_WPA3_SECURITY        0x01000000
+#define CYW43_AUTH_FLAG_ENTERPRISE_ENABLED   0x02000000
+#define CYW43_AUTH_FLAG_WPS_ENABLED          0x10000000
+#define CYW43_AUTH_FLAG_IBSS_ENABLED         0x20000000
+#define CYW43_AUTH_FLAG_FBT_ENABLED          0x40000000
+
 /**
  * \name Authorization types
  * \brief Used when setting up an access point, or connecting to an access point
  * \anchor CYW43_AUTH_
  */
 //!\{
-#define CYW43_AUTH_OPEN (0)                     ///< No authorisation required (open)
-#define CYW43_AUTH_WPA_TKIP_PSK   (0x00200002)  ///< WPA authorisation
-#define CYW43_AUTH_WPA2_AES_PSK   (0x00400004)  ///< WPA2 authorisation (preferred)
-#define CYW43_AUTH_WPA2_MIXED_PSK (0x00400006)  ///< WPA2/WPA mixed authorisation
-#define CYW43_AUTH_WPA3_SAE_AES_PSK  (0x01000004)   ///< WPA3 AES authorisation
-#define CYW43_AUTH_WPA3_WPA2_AES_PSK (0x01400004)   ///< WPA2/WPA3 authorisation
+//! No authorisation required (open)
+#define CYW43_AUTH_OPEN              (0)
+//! WEP PSK Security with open authentication
+#define CYW43_AUTH_WEP_PSK           (CYW43_AUTH_FLAG_WEP_ENABLED)
+//! WPA authorisation
+#define CYW43_AUTH_WPA_TKIP_PSK      (CYW43_AUTH_FLAG_WPA_SECURITY | CYW43_AUTH_FLAG_TKIP_ENABLED)
+//! WPA2 authorisation (preferred)
+#define CYW43_AUTH_WPA2_AES_PSK      (CYW43_AUTH_FLAG_WPA2_SECURITY | CYW43_AUTH_FLAG_AES_ENABLED)
+//! WPA2/WPA mixed authorisation
+#define CYW43_AUTH_WPA2_MIXED_PSK    (CYW43_AUTH_FLAG_WPA2_SECURITY | CYW43_AUTH_FLAG_AES_ENABLED | CYW43_AUTH_FLAG_TKIP_ENABLED)
+//! WPA3 AES authorisation
+#define CYW43_AUTH_WPA3_SAE_AES_PSK  (CYW43_AUTH_FLAG_WPA3_SECURITY | CYW43_AUTH_FLAG_AES_ENABLED)
+//! WPA2/WPA3 authorisation
+#define CYW43_AUTH_WPA3_WPA2_AES_PSK (CYW43_AUTH_FLAG_WPA3_SECURITY | CYW43_AUTH_FLAG_WPA2_SECURITY | CYW43_AUTH_FLAG_AES_ENABLED)
 //!\}
 
 /*!
@@ -221,9 +243,10 @@ typedef struct _cyw43_ev_scan_result_t {
     uint8_t ssid[32];   ///< wlan access point name
     uint32_t _2[5];
     uint16_t channel;   ///< wifi channel
-    uint16_t _3;
-    uint8_t auth_mode;  ///< wifi auth mode \ref CYW43_AUTH_
+    uint8_t _3[4];
     int16_t rssi;       ///< signal strength
+    uint8_t _4[12];
+    uint32_t auth_mode; ///< security and authentication flags
 } cyw43_ev_scan_result_t;
 //!\}
 
--- a/src/cyw43_ll.c
+++ b/src/cyw43_ll.c
@@ -263,6 +263,13 @@ static void cyw43_xxd(size_t len, const uint8_t *buf) {
 #define DOT11_IE_ID_RSN               (48)
 #define DOT11_IE_ID_VENDOR_SPECIFIC   (221)
 #define WPA_OUI_TYPE1                 "\x00\x50\xF2\x01"
+#define IEEE_802_11_OUI               "\x00\x0F\xAC"
+#define IEEE_802_11_OUI_LEN           3
+
+// Field offsets for authentication (RSN/WPA) IEs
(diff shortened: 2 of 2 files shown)


r-c-n:
Some notes:

- I could only test this on RSN-capable APs. Although the parsing of both RSN and WPA IEs should be similar, it'd be good to test on older APs as well to have proper validation.
- I couldn't test every authentication and security mode combination, but the more common ones (open, WPA2-PSK, WPA3-SAE) were tested ok.
- This possibly addresses issue https://github.com/georgerobotics/cyw43-driver/issues/141

src/cyw43_ll.h
```diff
@@ -222,7 +237,7 @@ typedef struct _cyw43_ev_scan_result_t {
     uint32_t _2[5];
     uint16_t channel;   ///< wifi channel
     uint16_t _3;
-    uint8_t auth_mode;  ///< wifi auth mode \ref CYW43_AUTH_
```
peterharperuk, line 233: Can you check this change doesn't break the rssi value? I seem to remember (my memory might be failing me) that the driver expects this structure to match the BSS info structure from the 43439 chip. auth_mode is actually uint8_t dtim_period, So changing this to uint32_t might trample over the rssi value (if my suspicions are correct)?

r-c-n: Good catch and thanks for reviewing! Yes, you're right. I just pushed an updated version that moves the `auth_mode` field to another place in the struct to keep the integrity of the data. I picked a 4-byte reserved region that shouldn't clash with anything currently used by the driver.

src/cyw43_ll.h
```diff
@@ -221,9 +236,10 @@ typedef struct _cyw43_ev_scan_result_t {
     uint8_t ssid[32];   ///< wlan access point name
     uint32_t _2[5];
     uint16_t channel;   ///< wifi channel
-    uint16_t _3;
-    uint8_t auth_mode;  ///< wifi auth mode \ref CYW43_AUTH_
+    uint8_t _3[3];
     int16_t rssi;       ///< signal strength
+    uint8_t _4[7];
```
peterharperuk: Similarly the [7] here looks odd. I would make it [12] as the 4th word is reserved and so definitely unused.
You would earn brownie points for defining everything properly - here's my effort...

// This is a "better" version of cyw43_ev_scan_result_t
typedef struct detailed_scan_result {
    uint32_t buf_len;
    uint32_t version;
    uint16_t sync_id;
    uint16_t bss_count;
    uint32_t bss_ver; // ie offset from here (12 bytes)
    uint32_t length;
    uint8_t bssid[6];
    uint16_t beacon_period;
    uint16_t capability;
    uint8_t ssid_len;
    uint8_t ssid[32];
    uint8_t pad1[1];
    uint32_t rates[5];
    uint16_t channel;
    uint16_t atim_win;
    uint8_t dtim_period; // was auth_mode
    uint8_t pad2[1];
    int16_t rssi;
    uint8_t phy_noise;
    uint8_t n_cap;
    uint8_t pad3[2];
    uint32_t nbss_cap;
    uint8_t ctl_ch;
    uint8_t pad4[3];
    uint32_t auth_mode; // using reserved word
    uint8_t flags;
    uint8_t reserved[3];
    uint32_t basic_mcs[4];
    uint16_t ie_offset; // offset of ie data from bss_ver
    uint8_t pad5[2];
    uint32_t ie_len;
    uint16_t snr;
} detailed_scan_result_t;

r-c-n: That's interesting, the struct you proposed seems to be incompatible to the reference I'm using. Here's the `cyw43_scan_result_internal_t` struct in `cyw43_ll.c`: https://github.com/georgerobotics/cyw43-driver/blob/055d64274b014dd7b1c2fc94d26e8a18face7124/src/cyw43_ll.c#L503

| offset | cyw43 driver                   |
|--------|--------------------------|
|      0 | version (4 bytes)        |
|      4 | length (4 bytes)         |
|      8 | bssid (6 bytes)          |
|     14 | beacon_period (2 bytes)  |
|     16 | capability (2 bytes)     |
|     18 | ssid_len (1 byte)        |
|     19 | ssid (32 bytes)          |
|     51 | rateset_count (4 bytes)  |
|     55 | rateset_rates (16 bytes) |
|     71 | chanspec (2 bytes)       |
|     73 | atim_window (2 bytes)    |
|     75 | dtim_period (1 byte)     |
|     76 | rssi (2 bytes)           |
|     78 | phy_noise (1 byte)       |
|     79 | n_cap (1 byte)           |
|     80 | nbss_cap (4 bytes)       |
|     84 | ctl_ch (1 byte)          |
|     85 | reserved32 (4 bytes)     |
|     89 | flags (1 byte)           |
|     90 | reserved (3 bytes)       |
|     93 | basic_mcs (16 bytes)     |
|    109 | ie_offset (2 bytes)      |
|    111 | ie_length (4 bytes)      |
|    115 | SNR (2 bytes)            |

Which matches the struct in the equivalent Linux driver: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/drivers/net/wireless/broadcom/brcm80211/brcmfmac/fwil_types.h?h=v7.1#n314

The struct you suggested has a slightly different layout starting at offset 51:

| offset | suggested                      |
|--------|--------------------------------|
|      0 | version (4 bytes)              |
|      4 | length (4 bytes)               |
|      8 | bssid (6 bytes)                |
|     14 | beacon_period (2 bytes)        |
|     16 | capability (2 bytes)           |
|     18 | ssid_len  (1 byte)             |
|     19 | ssid (32 bytes)                |
|     51 | pad1 (1 byte)                  |
|     52 | rates (20 bytes)               |
|     72 | channel (2 bytes)              |
|     74 | atim_win (2 bytes)             |
|     76 | dtim_period (1 byte)           |
|     77 | pad2 (1 byte)                  |
|     78 | rssi (2 bytes)                 |
|     80 | phy_noise (1 byte)             |
|     81 | n_cap (1 byte)                 |
|     82 | pad3 (2 bytes)                 |
|     84 | nbss_cap (4 bytes)             |
|     88 | ctl_ch (1 byte)                |
|     89 | pad4 (3 bytes)                 |
|     92 | auth_mode / reserved (4 bytes) |
|     96 | flags (1 byte)                 |
|     97 | reserved (3 bytes)             |
|    100 | basic_mcs (16 bytes)           |
|    116 | ie_offset (2 bytes)            |
|    118 | pad5 (2 bytes)                 |
|    120 | ie_len (4 bytes)               |
|    124 | snr (2 bytes)                  |

Which seems to be based on the Infineon WHD expansion driver: https://github.com/Infineon/whd-expansion/blob/0f13160971d02e7b1d4bd569da0d5eb96f2692d9/WHD/COMPONENT_WIFI5/inc/whd_types.h#L840

Maybe different versions of the wifi firmware provide different BSS info struct formats? I've only checked this driver with the provided firmware, have you tested it on more recent versions (newer than Version: 7.95.49)?

r-c-n: In any case, probably a cleaner solution would be to provide the scan result info in a separate driver struct instead of fitting additional data in unused parts of the buffer given by the hardware, although that change will have greater impact on the driver. For this particular use case, finding 4 unused bytes in the buffer will do.

peterharperuk: Even in your reference - I'd expect rateset_count  (a uint32_t?) to be have to be aligned on word boundary. So the compiler will add a byte of padding to move it to an offset of 52? (I could still be wrong of course).

r-c-n: Yeah, probably the paddings are there to account for field alignment... just checking now, the structs aren't packed. I'll push a fix. Thanks!

r-c-n: By the way, wrt defining a fully-detailed version of `cyw43_ev_scan_result_t`. The driver purposedly keeps a detailed `cyw43_scan_result_internal_t` version for parsing and publishes a reduced `cyw43_ev_scan_result_t` struct with only the fields relevant to consumers of the driver API.

I don't mind defining a single complete version as the public-facing struct, but in principle I'd keep the original intention of the driver and avoid adding all the bss info fields to the `cyw43_ev_scan_result_t` struct. Do you really prefer to provide all the fields in the public struct?

peterharperuk: It's probably better to keep things simple - I'm complicating things unnecessarily :)

src/cyw43_ll.h
```diff
@@ -161,18 +161,33 @@
...
+#define AES_ENABLED              0x0004
+#define SHARED_ENABLED       0x00008000
+#define WPA_SECURITY         0x00200000
+#define WPA2_SECURITY        0x00400000
+#define WPA2_SHA256_SECURITY 0x00800000
+#define WPA3_SECURITY        0x01000000
+#define ENTERPRISE_ENABLED   0x02000000
+#define WPS_ENABLED          0x10000000
+#define IBSS_ENABLED         0x20000000
+#define FBT_ENABLED          0x40000000
```
dpgeorge: Since these constants are defined in a header, would be best to prefix them with something that puts them in the CYW43 namespace.

How about `CYW43_AUTH_FLAG_xxx`?

r-c-n: Sounds good, let me do the changes.

r-c-n: Done

dpgeorge:
Please check CI: code formatting and build errors.

dpgeorge:
Still quite a few build errors.

r-c-n:
> Still quite a few build errors.

Sorry, yes, I noticed. I'm working on it, I shouldn't do things in a hurry...

dpgeorge:
Testing this, indeed the auth component of scan results is changed by this PR.  But changed in a good way!  It now matches the `CYW43_AUTH_FLAG_xxx` constants properly and you get complete information about the AP security mode.

I have seen the following bits being set:
- CYW43_AUTH_FLAG_WEP_ENABLED
- CYW43_AUTH_FLAG_TKIP_ENABLED
- CYW43_AUTH_FLAG_AES_ENABLED
- CYW43_AUTH_FLAG_WPA2_SECURITY
- CYW43_AUTH_FLAG_WPA2_SHA256_SECURITY
- CYW43_AUTH_FLAG_WPA3_SECURITY

dpgeorge (approved):

r-c-n:
> Testing this, indeed the auth component of scan results is changed by this PR. But changed in a good way! It now matches the `CYW43_AUTH_FLAG_xxx` constants properly and you get complete information about the AP security mode.
> 
> I have seen the following bits being set:
> 
>     * CYW43_AUTH_FLAG_WEP_ENABLED
> 
>     * CYW43_AUTH_FLAG_TKIP_ENABLED
> 
>     * CYW43_AUTH_FLAG_AES_ENABLED
> 
>     * CYW43_AUTH_FLAG_WPA2_SECURITY
> 
>     * CYW43_AUTH_FLAG_WPA2_SHA256_SECURITY
> 
>     * CYW43_AUTH_FLAG_WPA3_SECURITY

Thank you for testing it!

dpgeorge:
Thanks for your work on this @r-c-n .  Now merged!

This will definitely require a major version increase.

PR #150, opened 2026-06-10
src: Improve "bad auth" problems.

peterharperuk:
This is a speculative improvement for issues some users seem to see with connecting a Pico to an access point.
Today for some reason the WiFi environment in our office is terrible and I'm seeing a lot of connection problems.
These all show up as "auth failure" as if the WiFi credentials are wrong.
Ignoring "Timeout" and "No ack" status seems to improve reliability.

PR #154, opened 2026-07-16
src/cyw43_ctrl.c: Add bounded retry for failed auth.

Gadgetoid:
A Pico W connecting to a wireless network may fail to authenticate with a transient error and - in MicroPython at least - this failure surfaces to the user and requires a manual retry.

Handle auth failures internally with a bounded retry.

Turns the almost 100% failure to connect on first try (with my specific wireless network) to an almost 100% success rate.

### DIFF
--- a/src/cyw43.h
+++ b/src/cyw43.h
@@ -122,6 +122,9 @@ typedef struct _cyw43_t {
     bool pend_rejoin;
     bool pend_rejoin_wpa;
 
+    // Number of automatic rejoins remaining after a transient join failure.
+    uint8_t wifi_join_retries;
+
     // AP settings
     uint32_t ap_auth;
     uint8_t ap_channel;
--- a/src/cyw43_ctrl.c
+++ b/src/cyw43_ctrl.c
@@ -64,6 +64,15 @@
 #define WIFI_JOIN_STATE_KEYED   (0x0800)
 #define WIFI_JOIN_STATE_ALL     (0x0e01)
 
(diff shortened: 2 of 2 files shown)


dpgeorge:
I don't think I've ever seen this problem.

It's strange that the first attempt is failure yet the second attempt succeeds.  I assume that the AP is well in range?  Is is a repeatable problem?  Do you think it's specific to a particular router/AP?

I suspect the underlying cause is something else, although I wouldn't be against merging this workaround.  The downside is that a truly incorrect password now takes a lot longer to bubble up to the user as an error.

Gadgetoid:
> It's strange that the first attempt is failure yet the second attempt succeeds.

Very!

> Is is a repeatable problem? 

*Annoyingly* repeatable. It's a MikroTik L009 with 2.4GHz WiFi. It's always pretty much bulletproof once it connects.

> Do you think it's specific to a particular router/AP?

Extremely likely, but I don't know what specific issue with the router might be causing it. It seems to have persisted across multiple software updates. I've given to accept that flaky WiFi is a foregone conclusion, not even just on Pico.

> I suspect the underlying cause is something else

You're probably right, but I haven't the first clue how to figure out what it might specifically be.

peterharperuk:
@Gadgetoid I don't suppose you could run a test with a debug image with this PR? https://github.com/georgerobotics/cyw43-driver/pull/150
I added some logging of error codes we get from the firmware.

Gadgetoid:
Absolutely- but it’ll have to be next week. Currently zooming across Europe for IFA.

PR #155, opened 2026-08-13
src/cyw43_ll: call cyw43_write_bytes() with the real buffer size

r-c-n:
When `cyw43_sdpcm_send_common()` calls `cyw43_write_bytes()` to send a packet via SPI or SDIO, it accounts for the transport-specific padding when calling the function. That means that the underlying implementation of `cyw43_write_bytes()` receives the padded length and can't know the original buffer size.

[An implementation for SPI](https://github.com/georgerobotics/cyw43-driver/blob/055d64274b014dd7b1c2fc94d26e8a18face7124/src/cyw43_spi.c#L120) will use that padded length to build the SPI command, which will eventually cause the chip to emit the packet to the air with any trailing padding bytes added to it.

Padding the buffer should be a responsibility of the specific `cyw43_write_bytes()` implementation, and `cyw43_sdpcm_send_common()` should provide the original size of the buffer.

## Example:

Original buffer generated by the application (65 bytes):

```
0x20037640 <malloc_arena+43480>:        0x17    0xfe    0xfd    0x00    0x01    0x00    0x00    0x00
0x20037648 <malloc_arena+43488>:        0x00    0x00    0x13    0x00    0x34    0x00    0x01    0x00
0x20037650 <malloc_arena+43496>:        0x00    0x00    0x00    0x00    0x13    0xba    0x17    0x44
0x20037658 <malloc_arena+43504>:        0x83    0xb8    0xbc    0x2c    0x8f    0xeb    0x80    0x39
0x20037660 <malloc_arena+43512>:        0x1b    0x2d    0x3f    0x43    0x32    0x4c    0xc9    0x7a
0x20037668 <malloc_arena+43520>:        0x83    0x5e    0x8f    0xa6    0xfd    0x2a    0xb9    0x3b
0x20037670 <malloc_arena+43528>:        0x91    0x6f    0x01    0x81    0x14    0xf8    0xc9    0xb1
0x20037678 <malloc_arena+43536>:        0xb0    0x65    0x5b    0xfb    0x73    0x7a    0xac    0x90
0x20037680 <malloc_arena+43544>:        0x9d
```

By the time the buffer reaches `cyw43_sdpcm_send_common()`, it includes the additional UDP, IP and Ethernet headers, and `cyw43_sdpcm_send_common()` receives it as a 113 byte buffer. It then adds the SDPCM header (12 bytes). The resulting buffer is this (the payload starts at buf[18] = 0x90):

```
0x20017440 <cyw43_state+40>:    0x04    0x01    0xfb    0xfe    0xa6    0x02    0x00    0x76
0x20017448 <cyw43_state+48>:    0x00    0x85    0x00    0x00    0x00    0x00    0x20    0x00
0x20017450 <cyw43_state+56>:    0x00    0x00    0x90    0xde    0x80    0xde    0xf3    0x88
0x20017458 <cyw43_state+64>:    0x2c    0xcf    0x67    0xf2    0x07    0x3e    0x08    0x00
0x20017460 <cyw43_state+72>:    0x45    0x00    0x00    0x5d    0x00    0x00    0x00    0x00
0x20017468 <cyw43_state+80>:    0x40    0x11    0xae    0x29    0x0a    0x2a    0x00    0x6b
0x20017470 <cyw43_state+88>:    0xc0    0xa8    0x01    0x2a    0xf8    0xe3    0x99    0x0d
0x20017478 <cyw43_state+96>:    0x00    0x49    0x6d    0x51    0x17    0xfe    0xfd    0x00
0x20017480 <cyw43_state+104>:   0x01    0x00    0x00    0x00    0x00    0x00    0x13    0x00
0x20017488 <cyw43_state+112>:   0x34    0x00    0x01    0x00    0x00    0x00    0x00    0x00
0x20017490 <cyw43_state+120>:   0x13    0xba    0x17    0x44    0x83    0xb8    0xbc    0x2c
0x20017498 <cyw43_state+128>:   0x8f    0xeb    0x80    0x39    0x1b    0x2d    0x3f    0x43
0x200174a0 <cyw43_state+136>:   0x32    0x4c    0xc9    0x7a    0x83    0x5e    0x8f    0xa6
0x200174a8 <cyw43_state+144>:   0xfd    0x2a    0xb9    0x3b    0x91    0x6f    0x01    0x81
0x200174b0 <cyw43_state+152>:   0x14    0xf8    0xc9    0xb1    0xb0    0x65    0x5b    0xfb
0x200174b8 <cyw43_state+160>:   0x73    0x7a    0xac    0x90    0x9d
```

But then, when calling `cyw43_write_bytes()`, it'll pass the padded length (128) instead of the real buffer size (125), so `cyw43_write_bytes()` will build an SPI command (in this example) to send the whole buffer contents including the padding bytes:

```
0x20017440 <cyw43_state+40>:    0x7d    0x00    0x82    0xff    0x75    0x02    0x00    0x0e
0x20017448 <cyw43_state+48>:    0x00    0x00    0x00    0x00    0x00    0x00    0x20    0x00
0x20017450 <cyw43_state+56>:    0x00    0x00    0x90    0xde    0x80    0xde    0xf3    0x88
0x20017458 <cyw43_state+64>:    0x2c    0xcf    0x67    0xf2    0x07    0x3e    0x08    0x00
0x20017460 <cyw43_state+72>:    0x45    0x00    0x00    0x5d    0x00    0x00    0x00    0x00
0x20017468 <cyw43_state+80>:    0x40    0x11    0xae    0x29    0x0a    0x2a    0x00    0x6b
0x20017470 <cyw43_state+88>:    0xc0    0xa8    0x01    0x2a    0xf8    0xe3    0x99    0x0d
0x20017478 <cyw43_state+96>:    0x00    0x49    0x6d    0x51    0x17    0xfe    0xfd    0x00
0x20017480 <cyw43_state+104>:   0x01    0x00    0x00    0x00    0x00    0x00    0x13    0x00
0x20017488 <cyw43_state+112>:   0x34    0x00    0x01    0x00    0x00    0x00    0x00    0x00
0x20017490 <cyw43_state+120>:   0x13    0xba    0x17    0x44    0x83    0xb8    0xbc    0x2c
0x20017498 <cyw43_state+128>:   0x8f    0xeb    0x80    0x39    0x1b    0x2d    0x3f    0x43
0x200174a0 <cyw43_state+136>:   0x32    0x4c    0xc9    0x7a    0x83    0x5e    0x8f    0xa6
0x200174a8 <cyw43_state+144>:   0xfd    0x2a    0xb9    0x3b    0x91    0x6f    0x01    0x81
0x200174b0 <cyw43_state+152>:   0x14    0xf8    0xc9    0xb1    0xb0    0x65    0x5b    0xfb
0x200174b8 <cyw43_state+160>:   0x73    0x7a    0xac    0x90    0x9d    0x00    0x2c    0xcf
```

Here's the packet capture:

capture_cyw43_trailing_bytes

### DIFF
--- a/src/cyw43_ll.c
+++ b/src/cyw43_ll.c
@@ -712,8 +712,15 @@ static int cyw43_sdpcm_send_common(cyw43_int_t *self, uint32_t kind, size_t len,
 
     self->wwd_sdpcm_packet_transmit_sequence_number += 1;
 
+    #if CYW43_USE_SPI
+    // cyw43_write_bytes will pad the data to 4-byte alignment for the
+    // SPI transfer, but the actual payload size is encoded in the
+    // header word
+    return cyw43_write_bytes(self, WLAN_FUNCTION, 0, size, buf);
+    #else
     // padding is taken from junk at end of buffer
     return cyw43_write_bytes(self, WLAN_FUNCTION, 0, CYW43_WRITE_BYTES_PAD(size), buf);
+    #endif
 }
 
 struct ioctl_header_t {


dpgeorge:
Thanks for the patch.  This looks reasonable.

For SPI, `cyw43_write_bytes()` does already align the write to 4 bytes.  So this change should be OK there.

For SDIO, if the write is >64 bytes then `cyw43_sdio_cmd53()` will need to round up the length to 64 bytes, otherwise `sz /= block_size` will truncate data.  That will need fixing in this PR, and testing (I can test that).

r-c-n:
> Thanks for the patch. This looks reasonable.
> 
> For SPI, `cyw43_write_bytes()` does already align the write to 4 bytes. So this change should be OK there.
> 
> For SDIO, if the write is >64 bytes then `cyw43_sdio_cmd53()` will need to round up the length to 64 bytes, otherwise `sz /= block_size` will truncate data. That will need fixing in this PR, and testing (I can test that).

Thanks for reviewing and for the suggestion.

In order to make the SDIO change I need some information on the role of the `len` parameter of `cyw43_sdio_transfer_cmd53()`, though.

Say I pass `cyw43_sdio_cmd53()` the unpadded buffer length in `len`, and if it's larger than 64 bytes, then the `arg` parameter for `cyw43_sdio_transfer_cmd53()` is set up with block mode enabled and `sz` = number of blocks (padded length / block size). But then, in the `cyw43_sdio_transfer_cmd53()` call, is `len` supposed to still be the actual number of buffer bytes to transfer or is it expected to have the padding correction? The only SDIO implementation I can find is in [`mock_sdio.c`](https://github.com/georgerobotics/cyw43-driver/blob/36db3eecd3973b572a1efa4cce8bbf39bd10d610/tests/mock_sdio.c#L97), and it expects `len` to be either the same as the size encoded in `arg` if block mode is disabled, or to be the complete _padded_ length of the transfer if block mode is enabled. But if `len` can be derived from `arg`, why have it as a parameter?

dpgeorge:
I will need to do some investigation and testing, but I do know that on STM32 hardware, `block_size` must be a power of 2.  That means if `sz<=64` and not a power of 2, it will simply fail.

dpgeorge:
SPI is very different to SDIO in this context:
- SPI transfers are always byte oriented and the payload can be up to 2048 bytes.  The `arg` for SPI can specify the exact byte count of the payload yet it's still possible (and I think necessary) to round up the actual transfer length to 4 byte alignment.
- SDIO block transfers must be quantised to the `block_size` both in `arg` and in the actual transferred data amount.  So when using a block transfer it seems you cannot specify the exact byte size of the payload.

If you want to fast track this PR, then I suggest making the change only for when SPI is used.  For SDIO it should continue to pad to 64 bytes, and we can look at fixing/improving that later.

dpgeorge:
Testing SDIO transport on STM32F722 shows that (at least with byte mode, ie non-block mode) the number of bytes encoded in `arg` for CMD53 must match the number of bytes written to the SDIO bus (I tested only write mode).  And since the driver I'm testing with uses DMA in word mode, the length must be a multiple of 4 bytes.

So I'm not sure it's even possible to fix this issue when using the SDIO transport.

r-c-n:
> If you want to fast track this PR, then I suggest making the change only for when SPI is used. For SDIO it should continue to pad to 64 bytes, and we can look at fixing/improving that later.

Sounds good to me, I uploaded a new version. Thanks for testing!

src/cyw43_ll.c
```diff
@@ -712,8 +712,12 @@ static int cyw43_sdpcm_send_common(cyw43_int_t *self, uint32_t kind, size_t len,
 
     self->wwd_sdpcm_packet_transmit_sequence_number += 1;
 
+#if CYW43_USE_SPI
+    return cyw43_write_bytes(self, WLAN_FUNCTION, 0, size, buf);
```
dpgeorge, line 719: Could do with a comment, eg:
```c
// data on the SPI bus will be padded out to 4 byte-alignment by this function, but the true payload size is encoded in the header word
```

src/cyw43_ll.c
```diff
@@ -712,8 +712,12 @@ static int cyw43_sdpcm_send_common(cyw43_int_t *self, uint32_t kind, size_t len,
 
     self->wwd_sdpcm_packet_transmit_sequence_number += 1;
 
+#if CYW43_USE_SPI
```
dpgeorge: per CI, these #if/etc lines need indenting to match the code indent level

src/cyw43_ll.c
r-c-n: Done

src/cyw43_ll.c
r-c-n: Done

dpgeorge:
See #157 for a way to do this for the SDIO transport.

dpgeorge (approved):
Tested on RPi Pico W, and an Alif-based board using normal gSPI mode (not the hybrid RPi version of SPI).  All seems to work.

Let's get this merged, it's a good step forward.

PR #156, opened 2026-09-02
Change license of this driver to the MIT license.

dpgeorge:
This cyw43-driver has been around now for more than 4 years and in that time has had a semi-commercial license: free for non-commercial use, free for all uses in RP2xxx products, commercial for all else.

That license has served its purpose of generating revenue to fund the initial development.  The project is now better served with a fully open/free license, the MIT license.  Adopting the MIT license allows it to be more widely used, in particular within Zephyr.


dpgeorge:
I think it makes sense to accompany this license change with an increase in the major version number, namely moving to v2.0.0.

In that case it'd be good to merge some outstanding PRs first.  Then this one.  Then release v2.0.0.

PR #157, opened 2026-09-04
Improve SDIO CMD53 transfer size

dpgeorge:
This PR changes the signature of the back-end SDIO CM53 transfer function `cyw43_sdio_transfer_cmd53()`, to simplify it.  All the relevant information about write, block mode and length is encoded in the `arg` argument.

This now allows the back-end implementation to properly support byte-mode and block-mode for the transfers (prior to this change it was assumed that all transfers are done in block-mode, which is incorrect when that bit is unset in `arg`, although it still happened to work correctly).

CI is updated to pass the test.

Tested also on STM32F722 hardware.

Related to #155.


dpgeorge:
@r-c-n this is the best I could do with the SDIO driver.  It now pads only when necessary (when doing a block mode write).

r-c-n:
> @r-c-n this is the best I could do with the SDIO driver. It now pads only when necessary (when doing a block mode write).

If you tested it and it works, that's all I need to know, so you can replace https://github.com/georgerobotics/cyw43-driver/pull/155 with this one if you want. Thank you!

dpgeorge:
> If you tested it and it works, that's all I need to know, so you can replace #155 with this one if you want.

In order to keep changes clear and separated, I've merged #155.  This PR can now be done on top of that one.

dpgeorge:
I've updated this based on merge of #155, cleaned up the case of byte-mode with 512 bytes, and retested on hardware.

PR #159, opened 2026-09-08
Split the chipset blobs into separate firmware and CLM C-arrays

dpgeorge:
Chipset firmware and CLM data blobs are currently appended together (with padding to 512 bytes alignment) and stored as a single C array.

That's not necessary, they don't need to be one big array.  It was originally done so they could be stored separately to the firmware, but it's not helpful to do that anymore, and NVRAM and BT firmware are anyway separate.

Putting firmware and CLM in separate C-arrays makes it simpler to update to new firmware, eliminates unnecessary padding (reducing code size) and slightly simplifies the code that loads the firmware.  Also if we ever get back to implementing compression of the blobs, it should make that easier to do (see #68).

Users of this driver should not notice any difference, except possibly a bit smaller footprint due to eliminated padding.


dpgeorge:
I tested all 5 sets of affected firmware on appropriate hardware.

